PR #61, opened 2019-08-30
IS-7710: docs for gdpr-access-request API

tombech:


axelborgegmail:
Contect-info + type and Title is not required...

PR #108, opened 2020-05-28
IS-8661: added docs on the 'restore_completed' and 'pump_offset_set' notification rule types

knutj42:


knutj42:
NOTE: this PR should not be merged until the new notification rule types have been deployed to the production sesam portal (which will probably not happen until https://sesam.myjetbrains.com/youtrack/issue/IS-7758 has been resolved)

PR #150, opened 2021-01-18
IS-10402: Added 'Self-hosted service' document

grove:


self-hosted.rst
```diff
@@ -0,0 +1,170 @@
...
+===================
+
+.. contents:: Table of Contents
+   :depth: 2
+   :local:
+
+Introduction
+------------
+
+Sesam is offered as service that is either hosted for you in the cloud or you host it yourself. We generally recommend the former, but there are scenarios where you would want to host it yourself. This document explains the requirements and the installation steps necessary to get the single-machine Sesam service running in a self-hosted environment.
```
branislavjenco, line 12: Should we give an example under what scenario one might want to use self-hosted?

BaardBouvet (approved):

self-hosted.rst
BaardBouvet: One example would be that you are not allowed due to regulatory reasons to put data in the public cloud.

PR #188, opened 2021-03-15
Update core-principles.rst

JonasAls89:


BaardBouvet (approved):

msauban:
I think the section on the canonical model does not represent the intent accurately. While I agree it is an utopian goal to have an all encompassing eternal model, I do not believe it is the goal of the canonical model. I believe a good enough solution at a given time is achievable, I think the added value of Sesam's approach is on the flexibility front, and mechanically, on the cost front. The fact that a communication standard has to be redesigned every time a participating data format changes does not make the solution not viable, it makes it more expensive and prone to service unavailability.

PR #193, opened 2021-03-18
Update core-principles.rst

msauban:
suggestion for the definition of time series and reflexion on the canonical model

### DIFF
--- a/core-principles.rst
+++ b/core-principles.rst
@@ -54,20 +54,18 @@ architecture. <https://en.wikipedia.org/wiki/Service-oriented_architecture>`_
 The "simple" solution is to agree on one a single data model for each
 object type, a so-called `canonical
 model <https://en.wikipedia.org/wiki/Canonical_model>`_, that ideally
-represents a standardized superset everything you will ever need, from 
+represents a standardized superset everything you will need at the time, from 
 all systems. 
 While this may be a requirement in a message-based integration
 architecture, and is truly valuable in an industry-wide data 
 `interoperability <https://en.wikipedia.org/wiki/Interoperability>`_ 
 scenario, in a data platform however, the canonical model becomes a constant burden 
 demanding continual enhancement and expansion over time. The result becomes a 
-datamodel that fits none of the source systems, and always lags behind 
+datamodel that lags behind 
 the continously evolving needs of the platform.
 
-Many businesses have wasted enormous resources on a utopia of a stable canonical model,
-only to realize that it is not possible. An ideal model does not exist,
-and one must realize that any given data model will always be a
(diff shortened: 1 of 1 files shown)


msauban:
I think the section on the canonical model does not represent the intent accurately. While I agree it is an utopian goal to have an all encompassing eternal model, I do not believe it is the goal of the canonical model. I believe a good enough solution at a given time is achievable, I think the added value of Sesam's approach is on the flexibility front, and mechanically, on the cost front. The fact that a communication standard has to be redesigned every time a participating data format changes does not make the solution not viable, it makes it more expensive and prone to service unavailability.

axelborgegmail:
> I think the section on the canonical model does not represent the intent accurately. While I agree it is an utopian goal to have an all encompassing eternal model, I do not believe it is the goal of the canonical model. I believe a good enough solution at a given time is achievable, I think the added value of Sesam's approach is on the flexibility front, and mechanically, on the cost front. The fact that a communication standard has to be redesigned every time a participating data format changes does not make the solution not viable, it makes it more expensive and prone to service unavailability.

I was trying to include this in the current version of the text. Can you try to update the text so we can evaluate the actual change you want?

axelborgegmail:
> suggestion for the definition of time series and reflexion on the canonical model

I don't think I understand this change. A sensor reading represents a value in time, and I don't see in what case it can ever change?

msauban:
> > suggestion for the definition of time series and reflexion on the canonical model
> 
> I don't think I understand this change. A sensor reading represents a value in time, and I don't see in what case it can ever change?

I found the wording could lead to think the data never (ever) changes, when it can evolve through time.

msauban:
> > I think the section on the canonical model does not represent the intent accurately. While I agree it is an utopian goal to have an all encompassing eternal model, I do not believe it is the goal of the canonical model. I believe a good enough solution at a given time is achievable, I think the added value of Sesam's approach is on the flexibility front, and mechanically, on the cost front. The fact that a communication standard has to be redesigned every time a participating data format changes does not make the solution not viable, it makes it more expensive and prone to service unavailability.
> 
> I was trying to include this in the current version of the text. Can you try to update the text so we can evaluate the actual change you want?

I've updated the text. I removed notions of eternal solution.

msauban:
> > suggestion for the definition of time series and reflexion on the canonical model
> 
> I don't think I understand this change. A sensor reading represents a value in time, and I don't see in what case it can ever change?

I think the inclusion of "industry wide data sharing" vs "intra organization" is very good.

axelborgegmail (commented):
What about: "datamodel that is not native to any of the source systems, and lags behind

grove:
Decided not to merge this PR. Create a new one if you want.

PR #216, opened 2021-03-25
'#208: Adding Pipe-shortcuts 3.1'

JonasAls89:


ga-hegsvold (commented):
There are different shortcuts for Mac vs Windows/Linux. Maybe include both?

training/030_dtl/020_Beginner.rst
```diff
@@ -694,15 +694,40 @@ til "Restart"
...
+- Automatically formats your dtl.
+
+**Quick save**
+
+`Ctrl+s`
+
+- Saves the current pipe config.
+
+**Find/Replace**
+
```
ga-hegsvold, line 719: I think it would be good to also include the shortcut for Replace (Ctrl+h) and maybe add a remark that hitting Ctrl+f twice will also take you to the replace dialog.

gabriellsesam:
Good points from Gah here. I would also like to see a small reference to the "Editor Options" button on the Pipe config page as this will show the user all available shortcuts on the fly.

Screenshot 2021-03-26 at 07 22 36

gabriellsesam (changes requested):
See previous comments

JonasAls89:
ctrl+h does not seem to be working on MacOS GAH, so I didn't add it. Maybe we can discuss it during our next meeting.

training/030_dtl/020_Beginner.rst
ga-hegsvold: @JonasAls89 It's probably Cmd+h on MacOS. Could you please check and report it as a bug to support if it does not work?

I tried hitting Ctrl+f twice now to activate Search-Replace and that did not work. I've reported that to support.

ga-hegsvold (changes requested):
Just a follow-up on the Replace shortcut.

JonasAls89:
mh, no cmd+h minimises your browser window on MAC... 🤷‍♂️

ga-hegsvold (approved):

gabriellsesam (approved):

PR #221, opened 2021-03-26
#220: Edit: Architecture: Beginner: Datahub

gabriellsesam:


JonasAls89 (changes requested):
Agree with these changes, aside from lines 119-122, I would keep the lines I wrote.

JonasAls89 (approved):

training/010_architecture_and_concepts/020_Beginner.rst
```diff
@@ -103,45 +103,47 @@ Datahub @Jonas
...
 multiple different technologies, i.e., a data warehouse, microservices,
 databases etc.
 
 A Datahub shares data by connecting generators of data with consumers of
-data. In practice this allows data to be delivered in endpoints that can
-interact with the datahub in terms of injecting data into the datahub or
-exposing data to an external system. As such, a datahub mediates and
+data. In practice this allows data to be delivered in endpoints which
+interact with the datahub by either injecting into the datahub or
+exposing data to be retreived by the datahub. As such, a datahub mediates and
```
ga-hegsvold, line 114: I wonder if some semantics are lost here by rephrasing "an external system" to "be retreived by the datahub"?

The rephrasing seems to focus on the datahub being the receiver of data, either by push (inject) or pull (retreive).
In the next sentence we talk about mediation between systems so I think it's important here to point out that the datahub also makes data available for consumption.

OR...
We could skip the entire sentence starting with "In practice...". I think the paragraph still would convey the message.

ga-hegsvold (changes requested):

training/010_architecture_and_concepts/020_Beginner.rst
```diff
@@ -103,45 +103,47 @@ Datahub @Jonas
 
 Continuing the path of building data-centric solutions, the term
 “Datahub” comes into play. A datahub is recognized by its frictionless
-data flow and as such builds upon the architectural principles presented
+data flow and builds upon the architectural principles presented
 in DDA. A datahub can be described as a solution that consists of
 multiple different technologies, i.e., a data warehouse, microservices,
 databases etc.
 
 A Datahub shares data by connecting generators of data with consumers of
```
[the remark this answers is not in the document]
JonasAls89: Implemented.

training/010_architecture_and_concepts/020_Beginner.rst
JonasAls89: Deleted the sentence as suggested

ga-hegsvold (approved):

PR #237, opened 2021-04-12
IS-10915

Ole-Morten:
IS-10915


self-hosted.rst
```diff
@@ -11,6 +11,9 @@ Introduction
 
 Sesam is offered as service that is either hosted for you in the cloud or you host it yourself. We generally recommend the former, but there are scenarios where you would want to host it yourself. This document explains the requirements and the installation steps necessary to get the single-machine Sesam service running in a self-hosted environment.
 
+.. Note::
+   ``Self hosted subscriptions are required to run the latest image of their chosen `software channel <https://docs.sesam.io/concepts.html#software-channels>`_.
```
BaardBouvet, line 15: I think the link could be generated by sphinx, but then you need to test it by building it. This also works.

BaardBouvet (approved):

PR #256, opened 2021-04-20
#247 : adding generic inbound and outbound pipes section

JonasAls89:


training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -45,13 +45,33 @@ from multiple sources
 Coalesce, prioritization of source data (master data)
 
 
-.. _generic-input-pipes-custom-output-pipes-1-2:
+.. _generic-inbound-and-outbound-pipes-1-2:
 
-Generic input pipes, custom output pipes
+Generic inbound and outbound pipes
```
ga-hegsvold, line 50: A comment on the term "Generic": This term is a bit misleading, I think. The first thing that came to mind when I saw this title was that this is about pipes that can be reused for different purposes. Generic as opposed to specific (with a specific purpose). But that's not what is meant here. This is more about why we recommend doing "raw"/"pure" imports into Sesam and to not start changing or make assumptions about the data this early in the flow, and to minimize logic in endpoints.

I don't currently have a good substitute, but I do suggest we try to find a more precise term.

ga-hegsvold (changes requested):

training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -45,13 +45,33 @@ from multiple sources
...
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Write about where globals fit into the bigger picture of data flows, how
-do pipes going in look and how do pipes going out look?
+As established above, an important aspect when modelling data in Sesam is the use of globals. Albeit before reaching the global stage and after completion of the global stage, when modelling your data; - the following guidelines, respectively, apply.
```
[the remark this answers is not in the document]
JonasAls89: The first is the case. Therefore I chose modelling.

;- can be used to denote a pause, that is almost as long as a . 
I haven't got too strong opinions on this one, so I can change it.

training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -45,32 +45,33 @@ from multiple sources
...
+As data enters Sesam it is handled in inbound pipes. An inbound pipe should be as generic as possible with regards to the amount of shaping done on the data that flows through to its dataset. The reason being, in order for you to make the best possible modelling decisions downstream, you should look at the "raw" data first to get a complete understanding of the condition of the data. In addition, we want to assume as little as possible about how the data will be used by current and future recipients. Therefore,
+If we start shaping and customizing data too soon in the flow, it's much harder, if not impossible, to reuse the data for different purposes later. A rule of thumb is therefore to minimize the amount of DTL used in an inbound pipe and try to just copy everything, or close to everything. Special cases can occur when you need to do some shaping of the data before reaching the global stage. In such cases, you should aim at making the minimal required DTL changes in order for the data to retain as much of its original integrity as possible.
```
ga-hegsvold, line 59: Use lower case "if" here: "Therefore, If we" -> "Therefore, if we"

ga-hegsvold (changes requested):
Other than a small "typo" I think this is coming together nicely.

ga-hegsvold (approved):

gabriellsesam (approved):

PR #257, opened 2021-04-20
#251 : Adding - Filter entities on the way out

JonasAls89:


training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -58,18 +58,10 @@ do pipes going in look and how do pipes going out look?
...
 
-| Makes sure the endpoint only receives the entities they want.
-| Can stop entities from triggering events they shouldn’t trigger.
+Imagine you are working on a large dataset produced by a global pipe. You quickly recognize that the amount of data and all its properties is not that relevant to you. Therefore, one of the first things you do in your DTL is to filter on a specific key and value. This will leave you with a subsection of the complete data. Albeit, since the state of the data does not immediately satisfy your needs, you decide to add specific properties given a specific condition; i.e., if the entity is of type: "Employee" - add properties "Salary", "Position" and "Goals", albeit if it is not of type "Employee" add the property "NotRelevant". This written logic will provide us with a second means of filtering, namelig the property "NotRelevant". As illustrated, it is not unusual to use multiple filters in a DTL config, especially when the amount of DTL increases, and a need for stepwise filtering presents itself.
```
gabriellsesam, line 63: ´namelig` typo -> ´namely´

You should explicitly say that if the entity is not of type Employee it is therefore not relevant and can be filtered out. I think the "NotRelevant" property is confusing here because there is no mention of property's value.
```
if the entity is of type: "Employee" - add properties "Salary", "Position" and "Goals", albeit if it is not of type "Employee" add the property "NotRelevant". This written logic will provide us with a second means of filtering, namelig the property "NotRelevant".
```

training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -58,18 +58,10 @@ do pipes going in look and how do pipes going out look?
 Filter entities on the way out
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-| Filter gives the ability to stop entities from being sent by providing
-  a logical gate.
-| On the other hand, it can make sure you only send the entities you
-  wish to receive in an endpoint.
+Filtering entities after the global stage of modelling is a common use case. Filtering gives the ability to work with subsections of a dataset. It is therefore often times used when working on large datasets where you are only interested in a small section of the data. In addition, filtering is often times used in outbound pipes as well. This is due to the fact that *_deleted* entities never leave Sesam and is not usually something you would like to send to a target system. This is obviously not always the case, but in general that is how things tend to work.
```
gabriellsesam, line 61: `It is therefore often times used` -> `It is often used`

gabriellsesam: `filtering is often times used` -> `filtering is often used`

gabriellsesam: `*_deleted* entities never leave Sesam` -> `*_deleted* entities rarely leave Sesam`

erik-leven: subsections -> subsets

erik-leven: I don't agree. _deleted entities quite often leave Sesam and we often want them to. Maybe we don't have to explain the practical differance between _filetered and _deleted here, but since _deleted entities always are sent out of Sesam unless otherwise stated, I think this should come out in the text

JonasAls89: Good point that they are processed downstream yea. Albeit, when an _id is first introduced in Sesam, I don't think it ever leaves again, regardless of state of _deleted. I think Product talked about introducing a feature where you can set a time-limit within which entities exist as _deleted, but not sure it has been implemented yet.

JonasAls89: subsection: a division of a section.

subset: a part of a larger group of related things.

I think they are synonyms, so I assume we can use either one?

erik-leven: Well, then we should specify that _deleted entities can (and often are) sent out of Sesam, but they might not be "removed" from the datasets unless you yourself remove the dataset.

training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -58,18 +58,10 @@ do pipes going in look and how do pipes going out look?
...
 
-| Makes sure the endpoint only receives the entities they want.
-| Can stop entities from triggering events they shouldn’t trigger.
+Imagine you are working on a large dataset produced by a global pipe. You quickly recognize that the amount of data and all its properties is not that relevant to you. Therefore, one of the first things you do in your DTL is to filter on a specific key and value. This will leave you with a subsection of the complete data. Albeit, since the state of the data does not immediately satisfy your needs, you decide to add specific properties given a specific condition; i.e., if the entity is of type: "Employee" - add properties "Salary", "Position" and "Goals", albeit if it is not of type "Employee" apply a filter to exclude that entity. As illustrated, it is not unusual to use multiple filters in a DTL config, especially when the amount of DTL increases, and a need for stepwise filtering presents itself.
```
erik-leven, line 63: "Albeit, since the state of the data does not immediately satisfy your needs, you decide to add specific properties given a specific condition; i.e., if the entity is of type: "Employee" - add properties "Salary", "Position" and "Goals", albeit if it is not of type "Employee" apply a filter to exclude that entity."
Long sentence with two "albeit". Also, I'm not sure I follow the logic in the sentence. "since the state of the data does not..." -> this part was not intuitive for me

JonasAls89: noted. What word would you go with instead of "state"? or suggestion on rephrasing the first sentence?

JonasAls89: So far, looks like this: Albeit, since the state of the data does not immediately satisfy your needs, you decide to add specific properties given a specific condition; i.e., if the entity is of type: "Employee" - add properties "Salary", "Position" and "Goals". Finally, if it is not of type "Employee" apply a filter to exclude that entity.

erik-leven: Could we split these filter situations into two more clearly separate situations? The first one applies to the correct data coming into the pipe, the second applies to the quality of the data coming out of the pipe. Right now they are kinda mushed together

JonasAls89: alrigth

training/010_architecture_and_concepts/030_Novice.rst
ga-hegsvold: Regarding sections and sets: Considering we have "data**sets**" in Sesam, I think it's more precise to use "subset" than "subsection".
As a side note: for the dataset source there is a property called "subset" that only reads entities specified in the "subset" from the source.

JonasAls89: Think that is a valid argument. Correcting it.

training/010_architecture_and_concepts/030_Novice.rst
ga-hegsvold: > noted. What word would you go with instead of "state"? or suggestion on rephrasing the first sentence?

Maybe "content"? Or maybe just skip it: "since the data does not immediately satisfy your needs"

JonasAls89: Personally I do appreciate "state". It denotes the fact that we are shaping the data in our DTL and aligns well with the fact that data flows in states through pipes.

ga-hegsvold: I don't really have a strong opinion either way regarding "state". Just wanted to offer a couple of alternatives.
And to be clear: by "state of the data", I assume you mean which properties are currently present on the entities, right?

JonasAls89: That is right, yea.

gabriellsesam (approved):

training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -58,18 +58,10 @@ do pipes going in look and how do pipes going out look?
 Filter entities on the way out
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-| Filter gives the ability to stop entities from being sent by providing
-  a logical gate.
-| On the other hand, it can make sure you only send the entities you
-  wish to receive in an endpoint.
+Filtering entities after the global stage of modelling is a common use case. Filtering gives the ability to work with subsets of a dataset. It is therefore often used when working on large datasets where you are only interested in a small section of the data. In addition, filtering is often used in outbound pipes as well. This is due to the fact that *_deleted* entities are processed continously as data flows through Sesam and do rarely leave Sesam when first introduced. Additionally, *_deleted* entities are not usually something you would like to send to a target system. This is obviously not always the case, but in general that is how things tend to work.
```
ga-hegsvold, line 61: Is it sensible to use *_deleted* to reference entities with _deleted: true? Is this explained somewhere?
I don't mind us using it as long as it's explicitly explained prior to its usage.

ga-hegsvold (changes requested):

training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -58,18 +58,10 @@ do pipes going in look and how do pipes going out look?
...
 
-| Makes sure the endpoint only receives the entities they want.
-| Can stop entities from triggering events they shouldn’t trigger.
+Imagine you are working on a large dataset produced by a global pipe. You quickly recognize that the amount of data and all its properties is not that relevant to you. Therefore, one of the first things you do is to apply a filter on a specific key and value. This leaves you with a subset of the complete data. As you look closely at the state of the data, after having applied your first filter, you are not immediately satisfied. This makes you apply another filter to alter the state of the data further. Therefore, you decide to add a specific property given a specific condition; i.e., if the entity is of type: "Employee" - add properties "Salary", "Position" and "Goals". Finally, if it is not of type "Employee" apply a filter to exclude that entity. As illustrated, it is not unusual to use multiple filters in a DTL config, especially when the amount of DTL increases, and a need for stepwise filtering presents itself.
```
ga-hegsvold, line 63: Bare in mind I might be missing some context here, but reading only the "Filter entities on the way out" section, this example confuses me a bit.
Is there an example somewhere nearby of the data that is referred to in this paragraph?
Phrases like "You quickly recognize..." and "As you look closely..." would imply I'm actually looking at an example while reading this, no?

The order in which filters are step-wise added can of course vary, but I think it's sensible to point out that it's usually a good idea to start by filtering in relevant entities based on rdf:type(s).
Then you usually add any missing properties required by the target system by hopping to related datasets.
And lastly you can/should remove irrelevant properties.
During these steps you might also find that additional filters are required to further narrow down entities to those expected by the target system.

Thou, I would call adding and removing properties for "shaping" or "transforming", and not "filtering", so not sure if that is relevant to mention in this section.

ga-hegsvold (changes requested):

training/010_architecture_and_concepts/030_Novice.rst
JonasAls89: I feel like the above comment is a bit hard to address...

The example given is in theory, as this is in the architecture section of the book, I think it is proper to minimise the amount of DTL we use, as that should be used in the DTL section. Additionally, the phrasing used is in alignment with what we agreed upon, i.e. you are ... you should ... etc. Finally, the section starts off with "imagine you are working... " this implies that the coming example is in theory.

Again rdf:type, I believe, should be in the DTL section. If I introduce rdf:type now, I need to explain the concept first and then how we use rdf:type in Sesam DTL as well. I think this broadens the scope of this section too much.

I think your last comment, albeit I agree semantically, is more of a philosophical question. I don't think we should make it harder for the reader to consume. Therefore I think it is proper to not introduce words like shaping and transforming on top of filtering. This is for Novice, so I do think we should not assume too much about their knowledge..

Counter feedback?

training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -58,18 +58,10 @@ do pipes going in look and how do pipes going out look?
 Filter entities on the way out
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-| Filter gives the ability to stop entities from being sent by providing
-  a logical gate.
-| On the other hand, it can make sure you only send the entities you
-  wish to receive in an endpoint.
+Filtering entities after the global stage of modelling is a common use case. Filtering gives the ability to work with subsets of a dataset. It is therefore often used when working on large datasets where you are only interested in a small section of the data. In addition, filtering is often used in outbound pipes as well. This is due to the fact that *_deleted* entities are processed continously as data flows through Sesam and do rarely leave Sesam when first introduced. The *_deleted* property is used in SESAM to flag whether an entity is deleted or not. As such an entity which is deleted will have the property: ``{"_deleted": true},`` whilst an entity that is not deleted will have the property: ``{"_deleted": false}.`` Additionally, *_deleted* entities are not usually something you would like to send to a target system. This is obviously not always the case, but in general that is how things tend to work.
```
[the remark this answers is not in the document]
JonasAls89: corrected. 👍

training/010_architecture_and_concepts/030_Novice.rst
ga-hegsvold: @JonasAls89 
I see how this example can be tricky to write given it's in an early chapter.

My initial reaction was mainly about the phrasing, that it felt a bit.. strong (not sure if that word best describes it) for it to be a theoretical/hypothetical example. When I read "You quickly recognize..." my brain went "Do I? I don't see any data...". Since it's an imaginary example it might be phrased something like "Let's say you recognize..." or "Assume you recognize..." just to reinforce to the reader that this is indeed a hypothetical example.

I agree that undefined terms and concepts should not be written about here.

The point of this example, as I understand it, is to show that it's ok to use multiple filters, and I think that point is conveyed to the reader.

ga-hegsvold (approved):

PR #258, opened 2021-04-21
#239 : Adding Entities / JSON (Key-value pairs)

JonasAls89:


training/010_architecture_and_concepts/020_Beginner.rst
```diff
@@ -537,6 +537,17 @@ Examples end ref to 1.1.13
 Entities / JSON (Key-value pairs)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+As stated earlier in this section, a dataset consists of a list of entities and an entity is of the type JSON. In addition, the entity consists of key-value pairs. A key-value pair is two related data elements. A key is a constant and defines what that data element is concerned with, i.e., postCode, email, phoneNumber, etc. Meanwhile, the value provides contextual information for a specific key. This could look like the following:
```
gabriellsesam, line 540: An entity is a JSON dictionary with the key _id.

erik-leven, line 540: Maybe something like "..., a dataset consists of a list of entities. An entity is a JSON type dictionary containing a set of key-value pairs identified by its unique identifyer". ?

JonasAls89: Entities are defined in an earlier section, so therefore thought I shouldn't repeat it. ?

JonasAls89: Like it

JonasAls89: A little causious because they are defined earlier, but I'll add most of it.

gabriellsesam (approved):

ga-hegsvold (approved):

PR #259, opened 2021-04-21
Add high level setup

axelborgegmail:


setup-environment.rst
```diff
@@ -0,0 +1,72 @@
...
+
+The following subscriptions are necessary for a complete Sesam
+environment:
+
+-  Development: Every developer will need a small subscription, with no
+   SLA or backup
+
+-  CI-test: A small subscription, including Standard SLA
+
+-  Test: Same size as production, including standard SLA and Backup.
```
BaardBouvet, line 24: "Staging" might be a more precise term for those subscriptions.

setup-environment.rst
```diff
@@ -0,0 +1,72 @@
+Setting up a Sesam environment
+==============================
+
```
BaardBouvet, line 3: Should we add a disclaimer saying that this covers Cloud setups, and if you consider self hosting follow this link.

trondbbarstad: I agree with Baard. We should put in such a disclaimer

setup-environment.rst
trondbbarstad: Shouldn't we link to both Standard SLA and Enterprise SLA here? And perhaps a link to what describes the backup.

setup-environment.rst
```diff
@@ -0,0 +1,72 @@
...
+The recommended way to work with Sesam is to use test-driven
+development. This requires that the configuration include test data for
+all incoming data types. The test data allows developers to work
+separately on their own Sesame subscription, testing changes locally
+before propagating them to the test environment. A CI environment is
+also set up to automatically verify that the integrity of existing data
+is not violated before updated configuration is deployed.
+
+1. Set up a source control system
```
trondbbarstad, line 53: We should link to this after source control system: https://docs.sesam.io/project-workflow.html#set-up-version-control

setup-environment.rst
```diff
@@ -0,0 +1,72 @@
...
+
+2. Set up the CI subscription and connect it to your source control flow
+
```
trondbbarstad, line 56: Add: https://docs.sesam.io/project-workflow.html#set-up-automatic-ci-testing

setup-environment.rst
```diff
@@ -0,0 +1,72 @@
...
+3. **OPTIONAL**: Set up automatic deployment from source control
+
+4. Ensure every developer has set up their local test environment
```
trondbbarstad, line 59: Link to https://docs.sesam.io/project-workflow.html#set-up-local-tests ?

setup-environment.rst
```diff
@@ -0,0 +1,72 @@
...
+
+[*\* LIST OF STANRDARD PORTS \**]
+
+Development environment
+-----------------------
+
+The recommended way to work with Sesam is to use test-driven
+development. This requires that the configuration include test data for
+all incoming data types. The test data allows developers to work
+separately on their own Sesame subscription, testing changes locally
```
trondbbarstad, line 48: Sesame typo

setup-environment.rst
trondbbarstad: SLAs are not part of docs as of now. The question is shouldn't they? Both Standard and Enterprise?

trondbbarstad: And I agree on Test size. I have experienced myself how tedious it can be to wait several days for to rebuild a test from scratch(worst-case three weeks!). So in my mind having backup of test systems are really essential to support any fixing of test system (even without SLAs).

PR #264, opened 2021-04-27
IS-10326: Docs and changelog about disable/eager_load microservices

branislavjenco:


grove (approved):

configuration.rst
```diff
@@ -8895,6 +8904,12 @@ Properties
      - ``1800``
      -
 
+   * - ``eager_load``
+     - Boolean
+     - A false value will prevent the microservice from running unless there's at least one pipe connected to it. Overrides setting in :ref:`service metadata <service_metadata_global_defaults_eager_load_microservices>`. Only works in :ref:`clustered architecture <multi_arch>`.
```
BaardBouvet, line 8909: If we're going to mention clustered arch here, we also need to mention it on the global setting.

branislavjenco: True, added it now

BaardBouvet (approved):

PR #266, opened 2021-04-29
#241 : Adding Special Sesam Attributes section

JonasAls89:


training/010_architecture_and_concepts/020_Beginner.rst
```diff
@@ -557,11 +557,28 @@ Ref. 1.2.19, 3.2.14
 Special sesam attributes
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-Namespace
+Namespaces
+^^^^^^^^^^
+Namespaces are inspired by RDF (The Resource Description Framework). You'll see them in terms of namespaced identifiers - also called NIs. A NI is a special datatype defined in the entity data model. In essence they are a string consisting of two parts, the namespace and the identifier. "~:foo:bar" is an example. The ~: is the type part that tells you that it is a namespaced identifier. foo in this case is the namespace and bar is the identifier.
+
+Properties can also have namespaces, but here the ~: part is not used. global-person:fullname is an example of such a namespaced property. Namespaced properties are essential when merging to avoid naming collisions and to maintain provenance of the properties.
```
gabriellsesam, line 564: and to maintain provenance of the properties -> and to maintain the source of properties ?

training/010_architecture_and_concepts/020_Beginner.rst
```diff
@@ -557,11 +557,28 @@ Ref. 1.2.19, 3.2.14
...
+
+A namespaced identifier is a unique reference to an abstract thing. It is an identifer. In Sesam it is not a globally unique identifier, but it is a unique identifier inside one Sesam datahub. There are mechanisms in place for collapsing and expanding namespaced identifiers to globally unique identifiers on import and export. You use the functions ["make-ni"] or ["ni"] to create them.
+
 
 Rdf:type
+^^^^^^^^
+The RDF type is metadata used to relate data and give some semantic context. When used with a namespace, it keeps track of the origin of the data, as well as the business type. It is composed upon input and will be used to relate and filter like you would use a foreign key.
```
gabriellsesam, line 571: Should give an example or reference "¨~:foo:bar" above

JonasAls89: added this

training/010_architecture_and_concepts/020_Beginner.rst
JonasAls89: changed as almost suggested

gabriellsesam (approved):

training/010_architecture_and_concepts/020_Beginner.rst
ga-hegsvold: I think we need to clarify namespaces and namespaced identifiers a bit. This topic can be a bit tricky to grasp.

I'll give it a shot:

Namespaces in Sesam are primarily used on properties, and its main functions are to ensure uniqueness across sources and to maintain property provenance (as you explain).
`"global-person:fullname"` is an example of a namespaced property, where "global-person" is the namespace and "fullname" is the property name (or key?).

Namespaced identifiers (NIs) are identifiers (i.e. property values) given a namespace.
`"source:reference": "~:foo:bar"` is an example of a NI, where "source" is the property namespace, "reference" is the property name (key?), "foo" is the namespace of the references data and "bar" is the identifier (property value) matching an identifier in the referenced data. (maybe use an example that illustrates how a NI matches an identifier/$ids in another dataset?)
The "\~" is Sesam's way of telling you that it is of datatype NI. (this sentence should probably be improved)

NIs in Sesam are similar to foreign keys in databases in that NIs are a visual indication of how data is connected, and enables easier and more precise joins. However, Sesam does not enforce any relationship between NIs and the referenced properties.

ga-hegsvold (changes requested):

training/010_architecture_and_concepts/020_Beginner.rst
```diff
@@ -578,11 +578,32 @@ Ref. 1.2.19, 3.2.14
 Special sesam attributes
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-Namespace
+Namespaces
+^^^^^^^^^^
+Namespaces are inspired by RDF (The Resource Description Framework). You'll see them in terms of namespaced identifiers - also called NIs. A NI is a special datatype defined in the entity data model. In essence they are a string consisting of two parts, the namespace and the identifier. "~:foo:bar" is an example. The ~: is the type part that tells you that it is a namespaced identifier; foo in this case is the namespace and bar is the identifier.
```
ga-hegsvold, line 583: > Namespaces are inspired by RDF (The Resource Description Framework)

It may be true that the use of namespaces in Sesam was inspired by RDF, but the concept of namespaces have existed for many years and are used in many different places (Linux kernel, programming languages, XML, etc.).

Maybe we should add that "Namespaces *in Sesam* are inspired by RDF" if we need to reference RDF at all?

I think this paragraph overlaps, at least in part, with the NI explanation below.
I also think it would be easier for the reader to be introduced to the most common use of namepaces first (as property prefix), before diving into NIs.

ga-hegsvold (changes requested):

training/010_architecture_and_concepts/020_Beginner.rst
JonasAls89: so maybe just cut this paragraph out @ga-hegsvold ?

ga-hegsvold: Yeah, I think the remaining paragraphs cover namespaces and NIs sufficiently for beginner level.
And maybe cut "Properties can also have namespaces, but here the ~: part is not used." considering "~" will be explained further down.

ga-hegsvold (changes requested):

ga-hegsvold (approved):

PR #267, opened 2021-04-29
#243 : Adding NI for foreignKeys and make-ni examples

JonasAls89:


training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -19,6 +19,50 @@ Short overview of what data joining is
 Make namespaced identifiers for foreign keys - make-ni
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+Continuing along the lines of joining data, namespaced identifiers (NI's) come into play. NI's are used in Sesam to create unique identifiers inside one Sesam datahub. In order to create them, you can use either of the two functions: ["make-ni"] or ["ni"].
+
+In the below example, ["make-ni"] will be used. A NI in Sesam works like a foreign key in a relational database. As such, it shows the relation between data objects and enables the joining of these. The pipe config presented in the below example, shows exactly this:
```
erik-leven, line 24: Should we say why we need them? I mean, what we're doing here is basically just renaming a variable. Should we show an example of how we can actually use ne NI's in an equality, and maybe compare it to without a NI?

training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -19,6 +19,50 @@ Short overview of what data joining is
...
+	  "mssql-accounts:country": "DK",
+	  "mssql-accounts:id": 40,
+	  "mssql-accounts:phone": "1-894-115-3398",
+	  "mssql-accounts:phone-ni": "~:mssql-contacts:1-894-115-3398",
+	  "mssql-accounts:position": "CEO",
+	  "rdf:type": "~:mssql-accounts:accounts"
+	}
+
+
+As can be seen in the above dataset, the property with the key: "mssql-accounts:phone-ni" is the result of the function ["make-ni"] as defined in the above pipe config. The value can be used to join data between the pipes: "mssql-accounts" and "mssql-contacts" so that data can be merged to create complete representations of a related set of data objects. In Sesam, a merge is typically done on different datasets in the global stage of data modelling.
```
erik-leven, line 64: uten : etter key?

training/010_architecture_and_concepts/030_Novice.rst
JonasAls89: I think what you are suggesting in this comment, will be covered in other sections of this page. When we get a more complete picture, then maybe we need to revise some of it, but I think for now, I'll let it stay.

training/010_architecture_and_concepts/030_Novice.rst
JonasAls89: Endring implementeres

gabriellsesam (approved):

training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -19,6 +19,50 @@ Short overview of what data joining is
 Make namespaced identifiers for foreign keys - make-ni
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+Continuing along the lines of joining data, namespaced identifiers (NI's) come into play. NI's are used in Sesam to create unique identifiers inside one Sesam datahub. In order to create them, you can use either of the two functions: ["make-ni"] or ["ni"].
```
ga-hegsvold, line 22: typo: "NI's" -> "NIs"

> NI's are used in Sesam to create unique identifiers inside one Sesam datahub.

NIs are not really unique identifiers in Sesam, but are typically used to reference an identifier or a NI in a related dataset.

ga-hegsvold (changes requested):

ga-hegsvold (approved):

PR #279, opened 2021-05-12
High level planning

grove:


BaardBouvet (approved):

best-practices.rst
```diff
@@ -5,7 +5,7 @@ Best Practices
 .. toctree::
    :maxdepth: 2
 
+   Pre-planning a Sesam project <pre-planning>
```
branislavjenco, line 8: Is there a reason why "pre-planning" and not just "planning"?

branislavjenco: Also, is it okay if I fix a couple typos in a commit?

PR #287, opened 2021-05-21
#249 : Adding Customize data structure for outbound flow

JonasAls89:


training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -78,17 +78,46 @@ Tag your entities - Categorization of sub-concepts
 
 Extra:type - usually added into the globals to separate what entities about the same thing do & mean.
 
-.. _customize-data-structure-for-endpoints-1-2:
+.. _customize-data-structure-for-outbound-flow-1-2:
 
-Customize data structure for endpoints
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Customize data structure for outbound flow
```
gabriellsesam, line 83: "Customize data structure for outbound flow"
Verbally wrong, needs to be either:
`Customize data structure for an outbound flow`
or
`Customize data structure for outbound flows`

Could probably rename it entirely to f.ex
`Customizing data for outbound flows`

training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -78,17 +78,46 @@ Tag your entities - Categorization of sub-concepts
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-| Sesam has transformative functions to add, remove,Copy the attributes
-  you want the end system to receive.
-| All changes to attributes you add to the target will cause an entity
-  update.
+When data moves through Sesam in a dataflow from inbound towards the outbound stage of data modelling, it is typically necessary to transform your data so that it aligns to the schema your target system requires for consumption. Typical functions used when modelling data in the outbound stage could be: ``["add"], ["remove"], ["rename"], ["copy"].``
```
gabriellsesam, line 86: First half of that sentence is hard to read, should be rewritten like for example:
```
An outbound dataflow consists of the pipe which reads from a global, the pipe which sends to an external system and all the pipes in between.

In these outbound dataflows it is typically necessary to transform your data so that it aligns to the schema your target system requires for consumption.
```

training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -78,17 +78,46 @@ Tag your entities - Categorization of sub-concepts
...
 
-Referring to namespace 1.1.15 to know property origin, rename, add, copy
+As an example, the below data is produced by the global pipe named "global-person":
```
gabriellsesam, line 88: the below data is produced
->
the data below is produced

training/010_architecture_and_concepts/030_Novice.rst
JonasAls89: will implement.

training/010_architecture_and_concepts/030_Novice.rst
JonasAls89: will implement

training/010_architecture_and_concepts/030_Novice.rst
JonasAls89: I can rephrase, but I'm not sure I agree. To me the sentence is not wrong and not sure what you really mean about it being "verbally wrong"?

JonasAls89: I'll rephrase to this: Customize data structure for outbound flows
 
But would still appreciate an explanation.

training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -78,17 +78,46 @@ Tag your entities - Categorization of sub-concepts
...
+	{
+	  "global-person:country": "DK",
+	  "global-person:id": 40,
+	  "global-person:phone": "1-894-115-3398",
+	  "global-person:position": "Engineer",
+	  "mssql-accounts:positions": ["Engineer", "Salesmanager", "Accountant", "CTO"],
+	  "mssql-accounts:hobbies": "Builds LEGO"
+	}   
+
+The shape of the data does not immediately satisfy your needs, as you are only interested in working with the properties whose key starts with the namespace "global-person:". To solve this you choose to use the copy function where you can define what namespaces you are interested in. This would in DTL look like the following: ``["copy", "global-person:*"]`` and would produce the following data:
```
gabriellsesam, line 101: Would remove the first half of the first sentence, just replace it with
```
You are only interested in the properties with the namespace "global-person:"
```
________________________________
```
This would in DTL look like the following: ``["copy", "global-person:*"]`` and would produce the following data:
```
->
```
In DTL it would be written as ``["copy", "global-person:*"]`` and produce the following data:
```

training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -78,17 +78,46 @@ Tag your entities - Categorization of sub-concepts
...
+.. code-block:: json
+
+	{
+	  "global-person:country": "DK",
+	  "global-person:id": 40,
+	  "global-person:phone": "1-894-115-3398",
+	  "global-person:position": "Engineer"
+	} 
+
+After looking at the current shape and what your target system requires, you realise that you only need to pass on the properties "position", "id" and "phone". In addition, you recognize that the first letter of the keys must be in capital. To solve this in DTL, you would do the following: ``["remove", "country"]`` and ``["rename", "phone", "Phone"]``, ``["rename", "id", "Id"]``, ``["rename", "position", "Position"].`` Based on the declared DTL functions, this would produce the following:
```
gabriellsesam, line 112: I would like you to keep using the words you started this chapter with, specifically "schema".

```
After comparing the current shape of your data to the target system schema you realize only the properties "position", "id" and "phone" are needed.
```

gabriellsesam (changes requested):
I like the order you introduce the subject and how you use the examples, but I would like to see some changes for how the sentences themselves are structured and worded.

training/010_architecture_and_concepts/030_Novice.rst
gabriellsesam: It is just not correct english the way it is at the moment, you are missing an "an" or adding an "s" to the end of "flow"

training/010_architecture_and_concepts/030_Novice.rst
ga-hegsvold: @JonasAls89 quick comment:
"... aligns to the schema..." -> "... aligns with the schema..."

As a side note: The term "conform" is also a good substitute for "align".

PR #288, opened 2021-05-21
#249 : Adding corrections

JonasAls89:


training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -244,15 +244,44 @@ Extra:type - usually added into the globals to separate what entities about the
...
+
+.. code-block:: json
+
+	{
+	  "global-person:country": "DK",
+	  "global-person:id": 40,
+	  "global-person:phone": "1-894-115-3398",
+	  "global-person:position": "Engineer",
+	  "mssql-accounts:positions": ["Engineer", "Salesmanager", "Accountant", "CTO"],
+	  "mssql-accounts:hobbies": "Builds LEGO"
```
ga-hegsvold, line 262: We should avoid using database type ("mssql" in this case) as the system-part of the namespace in our examples.
Use the (fictional) name of the system instead.
Also use singular form for the table/datatype part of the namespace.
This is in accordance with our Best Practices.

So instead of "mssql-accounts", use for instance "crm-account" or whatever system name is appropriate for this example.

training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -244,15 +244,44 @@ Extra:type - usually added into the globals to separate what entities about the
...
+.. code-block:: json
+
+	{
+	  "global-person:country": "DK",
+	  "global-person:id": 40,
+	  "global-person:phone": "1-894-115-3398",
+	  "global-person:position": "Engineer"
+	} 
+
+After comparing the current shape of your data to the target system schema, you realize only the properties "position", "id" and "phone" are needed. In addition, you recognize that the first letter of the keys must be in capital. To solve this in DTL, you would do the following: ``["remove", "country"]`` and ``["rename", "phone", "Phone"]``, ``["rename", "id", "Id"]``, ``["rename", "position", "Position"].`` Based on the declared DTL functions, this would produce the following:
```
ga-hegsvold, line 276: This might be the OCD part of me talking, but I think it's a bit more pleasant to read if the properties that are referred to in the paragraph have the same order as they have in the example above:
"the properties "position", "id" and "phone" are needed" -> "the properties "id", "phone" and "position" are needed"

training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -244,15 +244,44 @@ Extra:type - usually added into the globals to separate what entities about the
...
 .. _customize-data-structure-for-endpoints-1-2:
 
-Customize data structure for endpoints
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Customize data structure for outbound flows
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+An outbound dataflow consists of the pipe which reads from a global pipe, the outbound pipe which sends data to an external system and all the pipes in between. In these outbound dataflows it is typically necessary to transform your data so that it aligns with the schema that your target system requires for consumption. Typical functions used when transforming data in the outbound stage could be: ``["add"], ["remove"], ["rename"], ["copy"].``
+
+As an example, the data presented below is produced by the global pipe named "global-person":
```
ga-hegsvold, line 252: This sentence could be simplified to:
As an example, the data presented below is produced by the pipe "global-person":

training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -244,15 +244,44 @@ Extra:type - usually added into the globals to separate what entities about the
 
 .. _customize-data-structure-for-endpoints-1-2:
 
-Customize data structure for endpoints
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Customize data structure for outbound flows
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+An outbound dataflow consists of the pipe which reads from a global pipe, the outbound pipe which sends data to an external system and all the pipes in between. In these outbound dataflows it is typically necessary to transform your data so that it aligns with the schema that your target system requires for consumption. Typical functions used when transforming data in the outbound stage could be: ``["add"], ["remove"], ["rename"], ["copy"].``
```
ga-hegsvold, line 250: Could this sentence:
> An outbound dataflow consists of the pipe which reads from a global pipe, the outbound pipe which sends data to an external system and all the pipes in between.

be simplified to:

> An _outbound_ dataflow consists of all pipes downstream from a global pipe.

maybe with emphasis on the term "outbound"?

ga-hegsvold (changes requested):
Overall I think this is good, but see my comments and questions.

training/010_architecture_and_concepts/030_Novice.rst
JonasAls89: as I've also got a little OCD I appreciate the logic in that. ;) will implement.

ga-hegsvold (changes requested):
I didn't notice on the first read, but I see that some of the function references are a bit difficult to read. I think we should prettify and make them stand out a bit more. See suggestions.

training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -244,15 +244,64 @@ Extra:type - usually added into the globals to separate what entities about the
...
+.. code-block:: json
+	
+	["remove", "country"] 
+
+and 
+
+.. code-block:: json
+	
+	["rename", "phone", "Phone"]
+	["rename", "id", "Id"]
```
ga-hegsvold, line 293: Would you mind listing these in alphabetical order as well (id, phone, position)? -OCD

ga-hegsvold (changes requested):
The new formatting looks good. Just one tiny detail. See below.

ga-hegsvold (approved):

gabriellsesam (approved):

Issue #294, opened 2021-06-03
Sesam In the Wild: unknown difficulty: Tag your entities - Categorize sub-concepts


gabriellsesam:
Entities in the same dataset which can have different meaning. 

Ex: Employees can have different Status based on attributes such as "isOnVacataion" or "Resigned" or if "resignedDate" is in the future or past. 

It can be tedious to have logic for such things in multiple places, so adding an attribute "sesam-employee:status" which can have the values "working", "not working" or "left" can be useful for filtering in downstream pipes.

PR #295, opened 2021-06-03
#294 : Adding this section to Novice in Sesam In tha wild

JonasAls89:


training/060_sesam_in_the_wild/030_Novice.rst
```diff
@@ -9,6 +9,88 @@ Sesam in the Wild: Novice
...
+
+The above description of the entity is a semantic categorization, as we define its meaning based on our understanding of the words that make up the entity. Albeit, from a semantic point of view, in terms of what can immediately be told from looking at the current state of the data, it seems that one of the properties can only be somewhat vaguely understood. Namely the property ``salesforce-person:EmployeeStatus.`` ``EmployeeStatus`` with the value ``0`` can mean many things. In such scenarios you might need to define sub-concepts that will allow you to increase the immediate semantic meaning. Imagine you are in charge of sending out salaries for the next month, albeit you are not sure whether or not Tordenskjold Skipskaptajnson is actually an active employee, and so should receive his registered salary. To solve this, you recognize that you need to enrich your data to increase the semantic meaning of the property ``salesforce-person:EmployeeStatus.`` This could for example be done by having a mapping dataset, which will contain a semantic description for a range of integers, i.e:
```
ga-hegsvold, line 37: The punctuation should be outside the formatting as it is not part of the Sesam config:
\``salesforce-person:EmployeeStatus.\`` --> \``salesforce-person:EmployeeStatus\``.

The combination of "This could" and "which will" in the sentence below doesn't feel quite right:

> This could for example be done by having a mapping dataset, which will contain a semantic description for a range of integers

I think it could be rephrased to:

> This could for example be done by having a mapping dataset containing a semantic description for a range of integers

training/060_sesam_in_the_wild/030_Novice.rst
```diff
@@ -9,6 +9,88 @@ Sesam in the Wild: Novice
...
+	  "EmployeeStatus": 3,
+	  "Description": "Future Employee"
+	}]  
+
+Then, if we use a ``hops`` function, we can merge the two datasets on ``EmployeeStatus.`` In DTL, this could look like the following:
+
+.. code-block:: json
+	
+	["merge",
+      ["hops", {
```
ga-hegsvold, line 63: Indentation is a bit off here..

training/060_sesam_in_the_wild/030_Novice.rst
```diff
@@ -9,6 +9,88 @@ Sesam in the Wild: Novice
...
+        "datasets": ["mapping-dataset md"],
+        "where": [
+          ["eq", "_S.salesforce-person:EmployeeStatus", "md.EmployeeStatus"] 
+        ],
+        "return": ["dict", "EmployeeStatusDescription", "md.Description"]
+      }]
+    ]
+
+And would produce the following changes to the dataset produced by the pipe ``salesforce-person:``
```
ga-hegsvold, line 72: Same case as with punctuation above:
\``salesforce-person:\`` --> \``salesforce-person\``:

training/060_sesam_in_the_wild/030_Novice.rst
```diff
@@ -9,6 +9,88 @@ Sesam in the Wild: Novice
...
+
+.. code-block:: json
+
+	{
+	  "_id": "salesforce-person:Tordenskjold Skipskaptajnson - Sesam Expert",
+	  "salesforce-person:Age": 32,
+	  "salesforce-person:CurrentProfession": "Sesam Expert",
+	  "salesforce-person:EmployeeStatus": 0,
+	  "salesforce-person:EmployeeStatusDescription": "On Leave",
```
ga-hegsvold, line 81: Wouldn't ``EmployeeStatusDescription`` technically get its namespace from the mapping-dataset here since it's a ``merge``?

training/060_sesam_in_the_wild/030_Novice.rst
```diff
@@ -9,6 +9,88 @@ Sesam in the Wild: Novice
...
+	  "salesforce-person:EmployeeTime": "5 years",
+	  "salesforce-person:FormerProfession": "Semiprofessional footballer",
+	  "salesforce-person:Fullname": "Tordenskjold Skipskaptajnson",
+	  "salesforce-person:Gender": "Male",
+	  "salesforce-person:Nickname": "Joe"
+	}  
+
+As can be seen from the above result, we are now able to tell that Tordenskjold Skipskaptajnson is currently "On Leave", which means that he should not be paid his monthly salary. In addition, this might also mean that he, against the odds, is again persuing a career in football.
```
ga-hegsvold, line 89: Funny! :-)

training/060_sesam_in_the_wild/030_Novice.rst
```diff
@@ -9,6 +9,88 @@ Sesam in the Wild: Novice
 Novice topic
 ~~~~~~~~~~~~
 
+
+.. _tag-your-entities-categorization-of-sub-concepts-6-2:
+
+Tag your entities - Categorization of sub-concepts
```
ga-hegsvold, line 15: I don't understand the "Tag your entities" part. What does that mean?

ga-hegsvold (changes requested):
Overall well written. The third paragraph is maybe a bit long. Consider breaking it up into a couple of smaller chunks if possible.

Besides that, see my comments.

training/060_sesam_in_the_wild/030_Novice.rst
JonasAls89: Not completely sure about the tag part either. But for now I'm gonna let it be.

training/060_sesam_in_the_wild/030_Novice.rst
ga-hegsvold: Line 64 should be dedented to match line 72.

training/060_sesam_in_the_wild/030_Novice.rst
ga-hegsvold: Anyone else got input on this?

ga-hegsvold (changes requested):

training/060_sesam_in_the_wild/030_Novice.rst
JonasAls89: fixed

training/060_sesam_in_the_wild/030_Novice.rst
gabriellsesam: Yea - when you have for example a table from your customer system called "cab-customer" and the data does not clearly define if the customer is government, private person or company then you can add logic either into the input pipe or the global which defines what "sub concept" the entities belong to, so you can easily filter in downstream pipes

gabriellsesam:
I like what you wrote here alot, and there is some overlap with the intent of this topic but not 100%.

I think we should rename this topic to "Mapping attributes across systems" - this would entail changing the case you've written here to include a second system which has it's own set of statuses and using the hop in a preparation pipe going to that third system. 

The mapping dataset should be called `sesam-map-salesforce-person-status` and the _id needs to be equivalent to the EmployeeStatus

training/060_sesam_in_the_wild/030_Novice.rst
```diff
@@ -9,6 +9,90 @@ Sesam in the Wild: Novice
 Novice topic
 ~~~~~~~~~~~~
 
+
+.. _tag-your-entities-categorization-of-sub-concepts-6-2:
```
gabriellsesam, line 13: Should be renamed to: Mapping the meaning of data across systems (?)

training/060_sesam_in_the_wild/030_Novice.rst
```diff
@@ -9,6 +9,90 @@ Sesam in the Wild: Novice
...
+
+The above description of the entity is a semantic categorization, as we define its meaning based on our understanding of the words that make up the entity. Albeit, from a semantic point of view, in terms of what can immediately be told from looking at the current state of the data, it seems that one of the properties can only be somewhat vaguely understood. Namely the property ``salesforce-person:EmployeeStatus``. ``EmployeeStatus`` with the value ``0`` can mean many things. In such scenarios you might need to define sub-concepts that will allow you to increase the immediate semantic meaning. 
+
+Imagine you are in charge of sending out salaries for the next month, albeit you are not sure whether or not Tordenskjold Skipskaptajnson is actually an active employee, and so should receive his registered salary. To solve this, you recognize that you need to enrich your data to increase the semantic meaning of the property ``salesforce-person:EmployeeStatus``. This could for example be done by having a mapping dataset containing a semantic description for a range of integers, i.e:
+
+.. code-block:: json
+	
+	[{
+	  "EmployeeStatus": 0,
```
gabriellsesam, line 44: Add: `"_id": "0",` to match the EmployeeStatus

training/060_sesam_in_the_wild/030_Novice.rst
```diff
@@ -9,6 +9,90 @@ Sesam in the Wild: Novice
...
+
+Usually an entity with its given properties will convey a concept. Imagine pulling data from a Customer relationship management (CRM) system. The inbound pipe that pulls data from this CRM system is called ``salesforce-person``. If you were to look at the entities in the dataset of this pipe, you will usually be able to understand what the data is all about. For brevity, only one entity is shown below:
+
+.. code-block:: json
+	
+	{
+	  "_id": "salesforce-person:Tordenskjold Skipskaptajnson - Sesam Expert",
+	  "salesforce-person:Age": 32,
+	  "salesforce-person:CurrentProfession": "Sesam Expert",
+	  "salesforce-person:EmployeeStatus": 0,
```
gabriellsesam, line 26: Add or explain a `make-ni` on the `employeeStatus` attribute
`["make-ni", "sesam-map-salesforce-person-status", "EmployeeStatus"]`

training/060_sesam_in_the_wild/030_Novice.rst
```diff
@@ -9,6 +9,90 @@ Sesam in the Wild: Novice
...
+
+Then, if we use a ``hops`` function, we can merge the two datasets on ``EmployeeStatus``. In DTL, this could look like the following:
+
+.. code-block:: json
+	
+	["merge",
+	    ["hops", {
+	      "datasets": ["mapping-dataset md"],
+	      "where": [
+		["eq", "_S.salesforce-person:EmployeeStatus", "md.EmployeeStatus"]
```
gabriellsesam, line 68: `["eq", "_S.salesforce-person:EmployeeStatus-ni", "md.$ids"]`

training/060_sesam_in_the_wild/030_Novice.rst
```diff
@@ -9,6 +9,90 @@ Sesam in the Wild: Novice
...
+	  "Description": "Future Employee"
+	}]  
+
+Then, if we use a ``hops`` function, we can merge the two datasets on ``EmployeeStatus``. In DTL, this could look like the following:
+
+.. code-block:: json
+	
+	["merge",
+	    ["hops", {
+	      "datasets": ["mapping-dataset md"],
```
gabriellsesam, line 66: `"datasets": ["sesam-map-salesforce-person-status"],`

training/060_sesam_in_the_wild/030_Novice.rst
```diff
@@ -9,6 +9,90 @@ Sesam in the Wild: Novice
...
+	      "where": [
+		["eq", "_S.salesforce-person:EmployeeStatus", "md.EmployeeStatus"] 
+	      ],
+	      "return": ["dict", "EmployeeStatusDescription", "md.Description"]
```
gabriellsesam, line 70: `"return": ["dict", "EmployeeStatusDescription", "md.sesam-map-salesforce-person-status:Description"]`

training/060_sesam_in_the_wild/030_Novice.rst
ga-hegsvold: @gabriellsesam this sounds like a sensible thing to do. Do you know of any general guidelines on how to do this?
Should we add new properties describing the "sub concepts"? Or maybe add new rdf:types?

In light of Gabriel's explanation of the title, is this text still on topic?
Even if it's not I do think we should keep it. Explaining how to use mapping datasets is very valuable, but maybe put it under another section? Something like "How to use mapping datasets"?

training/060_sesam_in_the_wild/030_Novice.rst
ga-hegsvold: @gabriellsesam 
I generally agree on this, but the "sesam" part of the mapping dataset feels a bit redundant. Is there a situation where we would prefix with something other than "sesam"? Maybe "map-salesforce-person-status" is sufficient?

ga-hegsvold (changes requested):
@gabriellsesam thanks for elaborating on the title and intention of this section.
I've added some questions and comments on your comments.

training/060_sesam_in_the_wild/030_Novice.rst
JonasAls89: I feel like it's a bit unintended to add a make-ni towards a mapping dataset...? I'll not add the "make-ni" until it's been clarified. To my knowledge the main purpose of "make_ni" is to mimic foreign-keys from relational databases. Adding it for a mapping dataset we create artificially feels a bit redundant to me. @ga-hegsvold , @gabriellsesam

training/060_sesam_in_the_wild/030_Novice.rst
JonasAls89: chose to go with: Creating mapping data for enrichment

training/060_sesam_in_the_wild/030_Novice.rst
ga-hegsvold: I view make-ni as a tool to help connect data. It's probably safe to say that in most cases it's used the way you describe, @JonasAls89, but I don't see anything wrong with using make-ni in other cases as well, as long as it eases the job of connecting data.
I've seen examples of database tables where no PKs or FKs where defined at all, but the intention was clearly to have certain fields operate as PKs and FKs. In this case it would make sense to use make-ni on the fields intended as FKs.

When it comes to mapping datasets, it can sometimes be beneficial to use make-ni, but not always.
In some cases it's not practical to use the _id / $ids as equality criteria. Sometimes it's much easier to just use values as-is without introducing namespaces in the property values.
There can also be cases where you want to reuse a mapping, but match on different properties in the mapping dataset.
In these cases I would not bother with using make-ni.

In this particular example I don't think it's strictly necessary to use make-ni.
You could argue that the status code mapping in this example could be reused by multiple sources.
So to map status from salesforce you could do lookup on a "salesforce-status-code" property in the mapping dataset.
To map from crm you could do lookup on a "crm-status-code" property, etc.
And in all cases you could return the "description" property.

training/060_sesam_in_the_wild/030_Novice.rst
ga-hegsvold: Sesam will actually give you $ids implicitly behind the scenes even if that property is not visible on the entity.
Since it's not obvious where the $ids property comes from here, we may not want to use it in this example.

ga-hegsvold (commented):
Add my thoughts.

ga-hegsvold:
Fixed merge conflict.

training/060_sesam_in_the_wild/030_Novice.rst
gabriellsesam: Your arguments make sense, make-ni in this instance would be a bit arbitrary

training/060_sesam_in_the_wild/030_Novice.rst
gabriellsesam: Yeah, in general i prefer hopping to the $ids as it makes indexing a non-issue, but to do so we need a pre-defined foreign key (which we wont do in this topic) so it is fine as is

gabriellsesam (approved):

training/060_sesam_in_the_wild/030_Novice.rst
```diff
@@ -9,6 +9,92 @@ Sesam in the Wild: Novice
...
+	  "salesforce-person:EmployeeStatus": 0,
+	  "salesforce-person:EmployeeTime": "5 years",
+	  "salesforce-person:FormerProfession": "Semiprofessional footballer",
+	  "salesforce-person:Fullname": "Tordenskjold Skipskaptajnson",
+	  "salesforce-person:Gender": "Male",
+	  "salesforce-person:Nickname": "Joe"
+
+	}  
+
+Which could be semantically defined as an object which is concerned with personal data. This definition is a very general one, as personal data can be many things. Meanwhile, for this specific entity, you could say that it is about the person Tordenskjold Skipskaptajnson whose age is 32 and who has been a Sesam Expert for 5 years. Previously, it seems he persued a career as a footballer, albeit he did not succeed in doing that given his current employment. This is a more detailed semantic definition of the above object.
```
ga-hegsvold, line 34: "Which could be" - what does "Which" refer to? Could it be re-written as "This entity could be"?

ga-hegsvold (changes requested):

ga-hegsvold (approved):

PR #305, opened 2021-06-17
#250 : Adding change tracking and data delta

JonasAls89:


training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -300,18 +300,11 @@ based on the declared DTL functions, this would produce the following:
...
-change in \_hash value a data-delta.
-
-Any data-delta for an entity in a dataset causes downstream pipes to see
-this as a new sequence number they haven’t yet read. This in turn makes
-the pipe process the entity. If the processed entity does not exist or
-gets a new \_hash in the output of the pipe, it will cause an update to
-the output dataset.
+Change tracking and data delta allows for Sesam to process and update data only when it changes. This ensures minimal latency and increases agility when 1)reading data from a source system and 2) when creating dataflows as data has entered Sesam and is modelled towards consumption in a target system.
+
+	1) When reading data from a source system, it may be possible to just ask for the data that have changed since the last time, if the source supports it. This mechanism uses entries from the source, such as a last updated time stamp, to ensure that only data that have been created, deleted or modified are processed.
```
ga-hegsvold, line 305: We can use # instead of hard-coded numbers to make dynamically numbered lists.
Indentation is also handled automatically.

---

It's a good practice to start with the condition, and then explain what happens if the condition is true (if this, then that).
Then we avoid surprising the readers with everything they read up to this point only applies if...

So suggested restructuring of this sentence:

> When reading data from a source system, it may be possible to just ask for the data that have changed since the last time, if the source supports it.

and put the "if ..."-part at the very start or right after the "When ..."-part.

training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -300,18 +300,11 @@ based on the declared DTL functions, this would produce the following:
...
-gets a new \_hash in the output of the pipe, it will cause an update to
-the output dataset.
+Change tracking and data delta allows for Sesam to process and update data only when it changes. This ensures minimal latency and increases agility when 1)reading data from a source system and 2) when creating dataflows as data has entered Sesam and is modelled towards consumption in a target system.
+
+	1) When reading data from a source system, it may be possible to just ask for the data that have changed since the last time, if the source supports it. This mechanism uses entries from the source, such as a last updated time stamp, to ensure that only data that have been created, deleted or modified are processed. 
+
+	2) Secondly, the first time data flows through a pipe in Sesam that pipe's dataset will be created. Datasets consist of entities and on each entity a ``hash`` property will be created. This ``hash`` property enables change tracking and data delta when data enters or flows through Sesam. When an entity's ``hash`` value changes, any downstream pipes register this change and recoqnizes it as a new sequence number that needs to be processed again.
```
ga-hegsvold, line 307: "Secondly" can be omitted since we already have a numbered list and the previous list item does not start with "Firstly".

---

\``hash\`` --> \``_hash\``

---
Suggestion:

> Datasets consist of entities and on each entity a \``hash\`` property will be created.

-->

> Datasets consist of entities and each entity has a \``_hash\`` property.

Should we explain which properties are used as base for the hashing or should that be covered in a higher learning tier?

ga-hegsvold (changes requested):
Mostly good. Some comments as always. ;-)

training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -300,18 +300,11 @@ based on the declared DTL functions, this would produce the following:
...
-time an entity is processed by a pipe. If the \_hash value changes or is
-new, the entity will be stored as a new version in dataset. We call this
-change in \_hash value a data-delta.
-
-Any data-delta for an entity in a dataset causes downstream pipes to see
-this as a new sequence number they haven’t yet read. This in turn makes
-the pipe process the entity. If the processed entity does not exist or
-gets a new \_hash in the output of the pipe, it will cause an update to
-the output dataset.
+Change tracking and data delta allows for Sesam to process and update data only when it changes. This ensures minimal latency and increases agility when 1)reading data from a source system and 2) when creating dataflows as data has entered Sesam and is modelled towards consumption in a target system.
```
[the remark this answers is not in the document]
JonasAls89: Implementing. with regards to numbering, that was the intent indeed.

training/010_architecture_and_concepts/030_Novice.rst
JonasAls89: I wasn't quite sure there, eventually I thought that would be a good subject down the road. @gabriellsesam  any thoughts?

training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -300,18 +300,11 @@ based on the declared DTL functions, this would produce the following:
...
-change in \_hash value a data-delta.
-
-Any data-delta for an entity in a dataset causes downstream pipes to see
-this as a new sequence number they haven’t yet read. This in turn makes
-the pipe process the entity. If the processed entity does not exist or
-gets a new \_hash in the output of the pipe, it will cause an update to
-the output dataset.
+Change tracking and data delta allows Sesam to process and update data only when it changes. This ensures minimal latency and increased agility both when importing data from source systems and when processing data through internal pipes towards target systems.
+
+Firstly, when reading data from a source system, if the source supports it, it may be possible to just ask for the data that have changed since the last time. This mechanism uses entries from the source, such as a last updated time stamp, to ensure that only data that have been created, deleted or modified are processed.
```
ga-hegsvold, line 305: Much better. Reading two "it"s in a row can break the flow a bit. To avoid this it can be rephrased from:
> if the source supports it

to:

> if supported by the source

training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -300,18 +300,11 @@ based on the declared DTL functions, this would produce the following:
...
-gets a new \_hash in the output of the pipe, it will cause an update to
-the output dataset.
+Change tracking and data delta allows Sesam to process and update data only when it changes. This ensures minimal latency and increased agility both when importing data from source systems and when processing data through internal pipes towards target systems.
+
+Firstly, when reading data from a source system, if the source supports it, it may be possible to just ask for the data that have changed since the last time. This mechanism uses entries from the source, such as a last updated time stamp, to ensure that only data that have been created, deleted or modified are processed. 
+
+Secondly, the first time data flows through a pipe in Sesam that pipe's dataset will be created. Datasets consist of entities and on each entity a ``_hash`` property will be created. This ``_hash`` property enables change tracking and data delta when data enters or flows through Sesam. When an entity's ``_hash`` value changes, any downstream pipes register this change and recoqnizes it as a new sequence number that needs to be processed again.
```
ga-hegsvold, line 307: typo: "recoqnizes" --> "recognizes"

ga-hegsvold (changes requested):
Small typo I missed on the first read and another suggestion.

ga-hegsvold (approved):
There are a couple of unresolved questions, but I'll leave it to others to comment further.

training/010_architecture_and_concepts/030_Novice.rst
gabriellsesam: I think we can just reference "Docs/Developer Guide/Entity Data Model"

gabriellsesam (approved):
See latest commit for minor changes, added see also section for _hash, change tracking etc

PR #335, opened 2021-08-25
#319 : Adding tasks for Beginners.

JonasAls89:
Not sure whether more tasks are needed for beginners?


training/050_microservices/020_Beginner.rst
```diff
@@ -295,4 +295,16 @@ Additional examples of external examples are listed below:
 Tasks for Microservices: Beginner – Tasks
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Run a microservice in Sesam [could be sink, http, source]
+#. *Why do we use microservices in Sesam?*
+
+#. *What is a microservice in Sesam?*
```
ga-hegsvold, line 300: Suggestion: Switch the order of the first two tasks so that the reader first defines what a microservice is, and then explains why it's used in Sesam.

ga-hegsvold (changes requested):
Good! Just a small suggestion.

ga-hegsvold (approved):
Nice! This is good start. More tasks can be added later if needed.

gabriellsesam (approved):

PR #374, opened 2021-10-04
#350 : Adding tasks

JonasAls89:


ga-hegsvold (approved):
I think this is good for now. The jury is still out regarding whether we should call "Developer" a group or a role.

PR #384, opened 2021-10-13
#336 : Adding initial commit of section

JonasAls89:


training/050_microservices/030_Novice.rst
```diff
@@ -10,31 +10,115 @@ Novice
...
+
+Finally, the topic of triggering comes into play. Triggering your system in Sesam is governed by how data flows in a given dataflow. An input pipe, that is connected directly to a system microservice, as shown in the above example, will run on a default schedule of 15 minutes, unless otherwise stated or if you choose to start the pipe manually. On the topic of run times, you can state specific run times by the use of cron expressions. A microservice that is used in the middle or at the end of a dataflow will be triggered when data flows through a specific pipe at this particular stage in the dataflow. As such, Sesam triggers microservices in accordance to the streams of data that makes a dataflow. A given change in the stream of data will trigger an interaction with a microservice, but only a load sufficient to handle the trigger will be carried out.
```
ga-hegsvold, line 107: I wonder if this:

> An input pipe, that is connected directly to a system microservice, as shown in the above example, will run on a default schedule of 15 minutes, unless otherwise stated or if you choose to start the pipe manually. On the topic of run times, you can state specific run times by the use of cron expressions.

can be compacted into this:

> An inbound pipe, as shown in the above example, will by default run every 15 minutes.

and that we add cross references to the explanation of inbound pipes and the use of cron expressions?

training/050_microservices/030_Novice.rst
```diff
@@ -10,31 +10,115 @@ Novice
...
-  TODO
+  Developer Guuide > JSON Pull Protocol: :ref:`json_pull_protocol`
+
+  Developer Guuide > JSON Push Protocol: :ref:`json_push_protocol`
+
+  Developer Guuide > Service Configuration: :ref:`continuation_support`
+
+  Learn Sesam: :ref:`systems-novice-2-2`
+
+  Developer Guuide > Service Configuration: :ref:`microservice_system`
```
ga-hegsvold, line 120: Please add full "bread crumbs", convert hard coded "bread crumbs" into ``:ref:``, and use ">" between all of them:

:ref:`developer-guide` > :ref:`json_pull_protocol`

:ref:`developer-guide` > :ref:`configuration` > :ref:`system_section` > :ref:`continuation_support`

etc..

PR #395, opened 2021-10-18
#337 : Adding initial repository

JonasAls89:


ga-hegsvold (dismissed):
See my specific comments below. Also a more general comment:

I would be reluctant to say that an IDE is a prerequisite. Most developers probably use some form of IDE, but it's quite possible to just use a text editor or some other basic tool.
We could recommend using an IDE, but I would leave it to the developer to decide whether to use it.
If we, in a later exercise, want to show specifically how to use an IDE for developing a MS, then I think it would be fine to say for that particular exercise this specific IDE is required.
But then we should also say which IDE to use.

Docker account I agree should be a prerequisite.

DockerHub account is only required if the developer wants to contribute to the Sesam Community. We can recommend it thou.

gabriellsesam (approved):

PR #471, opened 2021-11-24
#470 : Adding summary

JonasAls89:


ga-hegsvold (approved):
I think the summary points maybe should be cleaned up a bit, but don't have a better suggestion as of now so I'll approve the PR.

PR #472, opened 2021-11-24
#462 : Adding summaries and tasks

JonasAls89:
Also removed name mentions on titles and/or subsections, so Erik is not more.. 🪦


training/030_dtl/020_Beginner.rst
```diff
@@ -694,19 +721,27 @@ Result after run:
...
 
 DTL in practice
 ~~~~~~~~~~~~~~~
 
+.. sidebar:: Summary
+
+  DTL...
+
+  - is Sesam's language for shaping data
+
```
ga-hegsvold, line 736: I think the explanation of DTL belongs in another earlier section.
Maybe we could just use the bullet points blow as summary:

```
.. sidebar:: Summary

  In this section you will learn how to:

  - create a pipe from scratch
  - view the output of a pipe
  - write a greeting to the world with DTL
```

ga-hegsvold (changes requested):
One small suggestion. See comment.

ga-hegsvold (approved):

PR #477, opened 2021-11-26
#430 : Adding section

JonasAls89:


training/030_dtl/030_Novice.rst
```diff
@@ -72,11 +72,58 @@ As can be seen from the above produced data, the property with the key "_positio
...
+  }
+
+Which will produce the following dataset, when the pipe has completed a run:
+
+.. code-block:: json
+
+   {
+      "fakeKey": "fakeValue",
+      "fakeKey2": "fakeValue",
+      "newEmail": "thisIs@google.com",
```
ga-hegsvold, line 115: For this output to make sense we must also show the source data. Otherwise the value of `newEmail` pops out of nowhere.

There should be a paragraph between this result and the pipe config. Something like:

Given the following source data:
\<some sensible source data>
the above pipe config will produce the following output, ...:
\<this output>

training/030_dtl/030_Novice.rst
```diff
@@ -72,11 +72,58 @@ As can be seen from the above produced data, the property with the key "_positio
 "Add"
 ~~~~~
 
-Explain the add, based on ref 3.1.4 above
+.. sidebar:: Summary
+
+  "Add"...
+
+  - lets you define a new key and/or value
```
ga-hegsvold, line 79: "lets you define new properties" ?

ga-hegsvold (changes requested):

ga-hegsvold (approved):

PR #480, opened 2021-11-30
#431 : Adding section

JonasAls89:


training/030_dtl/030_Novice.rst
```diff
@@ -78,16 +78,87 @@ Explain the add, based on ref 3.1.4 above
...
+	        ]
+	     }
+	  },
+	  "metadata": {
+	    "global": true,
+	    "tags": "person"
+	  }
+  }
+
+When the above pipe runs, the following dataset will be produced:
```
ga-hegsvold, line 128: We need to show the source data here as well (same as for the section describing the "add" function). Or at least reference it if it's defined somewhere else.

ga-hegsvold (changes requested):

PR #485, opened 2021-12-02
#432 : Adding section

JonasAls89:
Recognized that we lacked some mention of RDF in architecture and concepts Novice. Also added additional information in architecture and concepts Beginner. I think this turned out better than initially outlined with regards to our explanation of RDF.


training/010_architecture_and_concepts/030_Novice.rst
```diff
@@ -93,6 +93,65 @@ As can be seen in the above dataset, the property with the key "mssql-accounts:p
 
   TODO
 
+.. _sesams-theoretical-approach-to-semantics-RDF-1-2:
+
+Sesams theoretical approach to semantics - RDF
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
ga-hegsvold, line 99: I'm not sure I understand the intention of this section.
The current content is essentially the same as this section: https://docs.sesam.io/rdf-support.html. We should avoid duplication.
What do we mean by "theoretical approach to semantics"?

training/030_dtl/030_Novice.rst
```diff
@@ -94,12 +94,61 @@ Concatenation of strings, examples etc
...
+	  "mssql-accounts:phone": "1-894-115-3398",
+	  "rdf:type": "~:mssql-accounts:accounts"
+	}
+
+As can be seen from the above pipe configuration, you have successfully added the RDF type property. The use of a namespace in this property allows for you to keep track of the origin of your entities, as well as retaining its business type. Therefore, it is composed upon input and will be used to relate and filter like you would use a foreign key.
 
 .. seealso::
 
-  TODO
+	:ref:`learn-sesam` > :ref:`architecture_and_concepts` > :ref:`architecture-and-concepts-novice-1-2` > :ref:`sesams-theoretical-approach-to-semantics-RDF-1-2`
```
ga-hegsvold, line 145: fix indentation

training/030_dtl/030_Novice.rst
```diff
@@ -94,12 +94,61 @@ Concatenation of strings, examples etc
...
+  - in Sesam could look like the following: ``{"rdf:type": "~:mssql-accounts:TheDanishKing"}``
+
+RDF stands for Resource Description Framework and is a standard for describing web resources and data interchange. Sesam has several features to facilitate working with RDF data both as `input <https://docs.sesam.io/rdf-support.html?highlight=rdf#rdf-input>`_, when doing transforms and finally when exposing or producing data for `external consumption <https://docs.sesam.io/rdf-support.html?highlight=rdf#rdf-output>`_.
+
+In DTL you will add an RDF property by doing the following:
+
+.. code-block:: json
+
+	{
+	  "_id": "mssql-accounts",
```
ga-hegsvold, line 112: Consider renaming the pipe `_id` to align more with Sesam's best practices: Avoid using database type (ex: mssql) as id, rather use database name (ex: agresso/salesforce/..).

ga-hegsvold (changes requested):

training/010_architecture_and_concepts/030_Novice.rst
JonasAls89: mhh.. Thought about that when adding it... Will try to contain myself to only new information, and then rather refer to what is already written in docs.

This is to ensure that a user will learn why our semantics (namespaces and NI's) look as they do. Also why it says theoretical, because this is the theory (URI, CURIE, N-Triples etc) that allows Sesam to implement the RDF standard.

JonasAls89: * that is the RDF standard, I should have said.

training/030_dtl/030_Novice.rst
```diff
@@ -262,12 +262,66 @@ As can be seen from the above dataset, you should recognize the properties with
...
+  ``rdf:type``...
+
+  - is applied in DTL by adding namespaces and namespaced identifiers
+  - is based on the source system id and the business type from that source system
+  - in Sesam could look like the following: ``{"rdf:type": "~:salesforce:Account"}``
+
+Resource Description Framework (RDF) is a standard for describing web resources and data interchange.
+Sesam has several features to facilitate working with RDF data both as `input <https://docs.sesam.io/rdf-support.html?highlight=rdf#rdf-input>`_,
+when doing transforms and finally when exposing or producing data for `external consumption <https://docs.sesam.io/rdf-support.html?highlight=rdf#rdf-output>`_.
+
```
ga-hegsvold, line 276: The RDF explanation is already covered in Architecture and Concepts. This section should focus on `rdf:type` specifically.
A few points I think should be explained here:
- What are rdf:types in Sesam?
- How do we decide on the value of rdf:types? (source system id + business type from source)
- At what step in a dataflow are rdf:types normally added? ("enrichment" step)
- Why add rdf:types? - i.e. what are they used for in Sesam? (easily identify source system and business type, suitable for filtering, more?)

I don't have good answers for all this, but I feel the use of rdf:types in Sesam needs a more in-depth explanation than what we currently have in the docs.

ga-hegsvold (changes requested):

training/030_dtl/030_Novice.rst
JonasAls89: I see your point here and is implementing, as I'm typing this, albeit I do very much agree we must include the steps defined in a sesam dataflow, I'm a little reluctant to start introducing it, enrichment, now, as this is opening up "pandoras box"... ;)

I'm not completely sure what has been decided with regards to the structure of the Sesambook, albeit I believe we should have these steps play a central part in it.

gabriellsesam (approved):
I don't fully understand Sesams use of RDF myself, but I find this topic nicely explains how it relates to namespaces and rdf:type.

Expanding it beyond these attributes may open up pandoras box as you say and make the topic less practical, so I think what's written here is good.

training/030_dtl/030_Novice.rst
ga-hegsvold: I see what you mean. Let's keep this section as it is for now then, but let's not forget that `rdf:type` and its purpose in Sesam should be explained in more detail somewhere in the docs.

ga-hegsvold (approved):

PR #486, opened 2021-12-07
#433 : Adding section

JonasAls89:


ga-hegsvold (approved):
Pipe id's should be updated according to Sesam's best practices, but that can be fixed later.

PR #487, opened 2021-12-07
#434 : Adding section

JonasAls89:


training/030_dtl/030_Novice.rst
```diff
@@ -119,11 +119,66 @@ EXAMPLESSS
...
-Declaraiton of foreign key in Sesam, explain /reference Namespace
+.. sidebar:: Summary
+
+  "Make-ni"...
+
+  - creates namespaced identfiers (NIs) by using the ``["make-ni"]`` function
+  - in Sesam is a complete Uniform Resource Locator (URL)
+  - is used to declare foreign keys as you would in a relational database management system (RDBMS)
+
+The ``["make-ni"]`` DTL function allows for defining `namespaced identifiers <https://docs.sesam.io/concepts.html?highlight=namespaced%20identifiers#namespaces>`_ (NIs). A NI in Sesam is a complete Uniform Resource Locator (URL). As such, it is used to investigate how data is semantically linked in a Sesam node. In addition, it is also used to declare foreign keys as you would in a relational database management system (RDBMS), albeit in Sesam the references will naturally be between pipes.
```
ga-hegsvold, line 130: I don't think we can say that NIs are complete URLs. If you copy-paste a NI into your browser it won't find much.
However, I think we can say that NIs can be **expanded** to complete **URI**s (https://en.wikipedia.org/wiki/Uniform_Resource_Identifier). A URI can be a URL (a locator) or a URN (just a name without location).

This ties in with the namespace prefixes that can be declared in a node's metadata.
Ex:
```
  "namespaces": {
    "default": {
      "owl": "http://www.w3.org/2002/07/owl#",
      "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
      "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
      "xml": "http://www.w3.org/XML/1998/namespace",
      "xsd": "http://www.w3.org/2001/XMLSchema#",
      "_": "https://foo.com/"
    }
```
So if my understanding is correct, a NI with for example the "rdf" namespace can be expanded to "http://www.w3.org/1999/02/22-rdf-syntax-ns#".

This gets technical real fast so we might be better off leaving out the URI here and rather explain this in a more advanced section.

I think overall our docs are a bit thin when it comes to explaining how all this works.

-----

"albeit in Sesam the references will naturally be between pipes."
Or rather between datasets, no?

training/030_dtl/030_Novice.rst
```diff
@@ -119,11 +119,66 @@ EXAMPLESSS
...
+
+As a NI is produced, after a pipe has completed its run, it will be prefixed with ``~:`` followed by the namespace and its value. To exemplify, look at the below example:
+
+.. code-block:: json
+
+	{
+	  "_id": "mssql-accounts",
```
ga-hegsvold, line 137: Can you update pipe id here as well and the corresponding namespaces in the output block below?

training/030_dtl/030_Novice.rst
```diff
@@ -119,11 +119,66 @@ EXAMPLESSS
...
+	  "mssql-accounts:id": 40,
+	  "mssql-accounts:phone": "1-894-115-3398",
+	  "mssql-accounts:phone-ni": "~:mssql-contacts:1-894-115-3398",
+	  "mssql-accounts:position": "CEO",
+	  "rdf:type": "~:mssql-accounts:accounts"
+	}
+
+As can be seen from the above output, the property ``"mssql-accounts:phone-ni"`` is the namespace that tells you that this is your recently created NI. The value of your NI is in practice your foreign key and tells you that the value of "phone" is a reference to the pipe named ``"mssql-contacts"``.
+
+Finally, as mentioned initially, the NI is in reality a URL, and as such you can press your NIs and navigate your Sesam node with respect to how data is semantically linked in your node.
```
ga-hegsvold, line 173: See comments above regarding URI/URL/URN.

ga-hegsvold (changes requested):

training/030_dtl/030_Novice.rst
```diff
@@ -119,11 +119,66 @@ EXAMPLESSS
...
+	    "system": "sesam-training",
+	    "table": "accounts"
+	  },
+	  "transform": {
+	    "type": "dtl",
+	    "rules": {
+	      "default": [
+	        ["copy", "*"],
+	        ["add", "rdf:type",
+	          ["ni", "salesforce-accounts", "accounts"]
```
ga-hegsvold, line 150: Didn't notice this eariler, but the rdf:type should just reference the source system + object type.
In this instance "salesforce" is the source system and "Account" could be a fitting object type, so I suggest we use the following:
`["add", "rdf:type", ["ni", "salesforce", "Account"]]`
which will give the following:
`"rdf:type": "~:salesforce:Account"`

training/030_dtl/030_Novice.rst
```diff
@@ -119,11 +119,66 @@ EXAMPLESSS
...
+
+.. code-block:: json
+
+	{
+	  "salesforce-accounts:country": "DK",
+	  "salesforce-accounts:id": 40,
+	  "salesforce-accounts:phone": "1-894-115-3398",
+	  "salesforce-accounts:phone-ni": "~:salesforce-contacts:1-894-115-3398",
+	  "salesforce-accounts:position": "CEO",
+	  "rdf:type": "~:salesforce-accounts:accounts"
```
ga-hegsvold, line 168: See rdf:type comment above.

ga-hegsvold (changes requested):

ga-hegsvold (approved):

PR #505, opened 2021-12-15
#439 : Adding section

JonasAls89:


training/030_dtl/030_Novice.rst
```diff
@@ -198,11 +198,76 @@ A: [{"foo":1},{"foo":2}] -> \_S.A.foo = [1,2]
...
+.. code-block:: json
+
+	{
+    "_id": "arcgis-grid-measure:1",
+    "arcgis-grid-measure:id": 1,
+    "arcgis-grid-measure:newTemperature": "50 degrees",
+    "arcgis-grid-measure:lastSensorID": 2
+  }
+
+Starting from the top, the source dataset ``"My_list"`` is a list with nested dictionaries. Such a data object is an ideal candidate for use in a nested transformation, such as an ``["apply"]`` function, shown in the "Pipe transform statement". As you should recognize, the ``["apply"]`` is used to access the nested elements in the source property ``"My_list"``. In addition, the function called ``"custom_function"`` is referenced and then applied when the ``["apply"]`` funcion evaluates. In order for the result to become part of the dataset, you will need to also merge the result to the default rule, which is why we need the ``["merge"]`` wrapper.
```
ga-hegsvold, line 264: > In addition, the function called ``"custom_function"``

-->

> In addition, the rule called ``"custom_function"``

I think we should rename the rule to something other than "_function" to avoid confusion with Sesam's actual functions. ``custom_rule`` would be fine.

training/030_dtl/030_Novice.rst
```diff
@@ -198,11 +198,76 @@ A: [{"foo":1},{"foo":2}] -> \_S.A.foo = [1,2]
...
+
+Going into detail with respect to what happens in our ``"custom_function"``, you could state, that applying it is like using a for loop, in a programming language, to send in entries from a list to a function. As such, our ``"custom_function"`` will be called for each index in the list that we pass onto it. This is also why ``"lastSensorID"`` evaluates to ``2``, since the last entry we pass onto it equals ``{"foo": 2}``. The if statement that is applied in our ``"custom_function"`` is an example of how the logic can disregard list indexes and rather evaluate property values. This is a useful strategy you can apply to further making use of the ``["apply"]`` function, in addition to extending your transform capabilities.
```
ga-hegsvold, line 266: > ``"custom_function"`` will be called for each index in the list

I think replacing "index" with "item" or "element" would be preferable here.
"ITEM" and "element" seems to be used interchangeably in the DTL Guide.
"INDEX" is the position (represented as an integer) of an item/element in a list.

ga-hegsvold, line 266: Where does ``{"foo": 2}`` come from? ;-)

ga-hegsvold (changes requested):

ga-hegsvold (approved):

PR #507, opened 2021-12-15
440

JonasAls89:


ga-hegsvold (approved):
Added the difference between merge and merge-union as bullet points to the Summary so that the reader can spot it at a glance.

PR #510, opened 2021-12-17
Cleaning the docs

JonasAls89:
Due to overlap and inconsistencies in this section compared to what we state in Best Practices, I removed this file.


BaardBouvet (approved):

BaardBouvet (changes requested):
Make sure we have included all relevant info from data modelling in other places before we merge this.

PR #516, opened 2021-12-21
#443 : Adding section

JonasAls89:


training/030_dtl/030_Novice.rst
```diff
@@ -244,51 +244,90 @@ Basics, uten apply
...
+	  "salesforce-person:employmentID": 40,
+	  "salesforce-person:name": "Bolton, Aladdin T.",
+	  "erp-person:ID": 40,
+	  "erp-person:employmentPosition": "CTO",
+	  "erp-person:phone": "1-894-115-3398",
+	  "rdf:type": [
+	    "~:salesforce-person:person",
+	    "~:erp-person:person"
+	  ],
+	  "erp-person:updateDate": "~t2021-12-21T10:59:39.94248192Z"
```
gabriellsesam, line 277: I would like this topic to mention the prefix we use for types of data, ~t for datetime, ~f for float

gabriellsesam (changes requested):
Good example, but missing explanation for the value prefix `~t`

gabriellsesam (approved):

PR #538, opened 2022-01-20
#437 : Adding concatenation

JonasAls89:


training/030_dtl/030_Novice.rst
```diff
@@ -621,16 +589,111 @@ As you can see from the above merged result, the ``_id`` turned out as ``"salesf
...
+
+.. sidebar:: Summary
+
+  "Coalesce"...
 
-ref 1.2.19
+  - is one of Sesam's core Master Data Management (MDM) capabilities
+  - lets you define a list of prioritized properties
+
+Coalesce is one of Sesam's core Master Data Management (MDM) capabilities. ``["coalesce"]`` lets you define a list of prioritized properties that will be evaluated so that you can make sure the first property that does not return ``null`` becomes the value of the property you are working on. An example has been drafted below to exemplify the use of ``["coalesce"]``.
```
gabriellsesam, line 604: Succinct and well written, but I would like the text to mention that "coalesce" supports custom expressions which can be useful when a null-check is not precise enough.

Example of a coalesce to get the first non-empty string:
```
["coalesce", 
  ["and",
    ["is-string", "_."], 
    ["neq", "_.", ""]
  ], 
  ["list", null, "", 999, "pickme"]
]

**Review by gabriellsesam — CHANGES_REQUESTED** [review]
All in all very good, but I'd like to see a couple of details mentioned for concat and coalesce

File: training/030_dtl/030_Novice.rst [hunk 2]
```diff
@@ -151,110 +151,78 @@ Which will produce the following output, when the pipe has completed a run:
...
-  "Coalesce"...
+  "Concat"...
 
-  - is one of Sesam's core Master Data Management (MDM) capabilities
-  - lets you define a list of prioritized properties
+  - is used when you need to concatenate multiple properties into one property with a desired shape
+  - can only evaluate string values
 
-Coalesce is one of Sesam's core Master Data Management (MDM) capabilities. ``["coalesce"]`` lets you define a list of prioritized properties that will be evaluated so that you can make sure the first property that does not return ``null`` becomes the value of the property you are working on. An example has been drafted below to exemplify the use of ``["coalesce"]``.
+Concatenation is used when you need to concatenate multiple properties into one property with a desired shape. In Sesam concatenation is applied with the DTL function ["concat"]. Important to know here, is that ["concat"] can only evaluate string values. Therefore, you need to make sure that all provided properties in the ["concat"] function are string data types. To exemplify using concatenation in Sesam look to the below example.
```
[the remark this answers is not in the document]
JonasAls89: Struggling a little to acknowledge when a concat will be used with a list..? As in an apply rule, where you concat each element in a list?.. @gabriellsesam

training/030_dtl/030_Novice.rst
JonasAls89: Thanks Gabbe. 👍 I think this example should be used in advanced, albeit I'm not sure how relevant this use of coalesce is... I feel coalesce should be used for master data management and the ranking of namespaces as outlined here. I am feeling the above might not be that relevant for use when thinking about what coalesce should be used for. Albeit, we can get back to this when we write coalesce for advanced, if it comes up then. Ok?

gabriellsesam: > Thanks Gabbe. 👍 I think this example should be used in advanced, albeit I'm not sure how relevant this use of coalesce is... I feel coalesce should be used for master data management and the ranking of namespaces as outlined here. I am feeling the above might not be that relevant for use when thinking about what coalesce should be used for. Albeit, we can get back to this when we write coalesce for advanced, if it comes up then. Ok?

You're right, more important to get the point across then overwhelm the reader

training/030_dtl/030_Novice.rst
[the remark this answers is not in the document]
gabriellsesam: Hmm fair point, I usually write concats like…
`["concat", ["list", "a", "b, "c"]`
but there is not really a good reason to do so, so no value added by mentioning it.

gabriellsesam (approved):

PR #569, opened 2022-02-07
IS-11682: Document change in growth_threshold

hglad:


configuration.rst
```diff
@@ -677,9 +677,9 @@ Properties
    * - ``compaction.growth_threshold``
      - Float
      - The growth factor required for the automatically scheduled compaction to kick
-       in. The default value is that there must have been 10% new offsets written to
-       the dataset since the last compaction. ``1.0`` is the minimum value allowed.
-     - ``1.10``
+       in. Uses the minimum value of ``1.0`` by default, meaning that no new offsets
+       need to be written to the dataset for compaction to run.
```
grove, line 681: Compaction won't run when there are no changes to the dataset. I would change this to: "meaning that compaction will always run when new entities are written to the dataset".

grove (approved):

PR #571, opened 2022-02-07
Adding github autodeployer and /demo folder for CI/CD workflow

JonasAls89:


BaardBouvet (changes requested):
No need for the "node" subdirectory? If we think "demo" is not descriptive enough, I suggest "demo-node" or something like that.

BaardBouvet (changes requested):
Sorry about suggesting "node" in the folder name. We discussed this word internally and really don't want to use it anymore. How about "demo-config"? That should stand the test of time.

BaardBouvet (approved):

PR #594, opened 2022-03-03
Update patterns.rst

JonasAls89:


patterns.rst
```diff
@@ -101,6 +101,14 @@ Optimistic locking
 ------------------
 Should be added via an external transform and then two hash values should be compared. In case of difference, discard entity.
 
+Subsets
```
BaardBouvet, line 104: I think we should call it "Only process the relevant entities from a global using subset"

patterns.rst
```diff
@@ -101,6 +101,14 @@ Optimistic locking
...
+-------
+If your source supports it, you should use the :ref:`subset <dataset_source>` property to pick subsets of a complete dataset, i.e: after a merge source has been used upstream.
```
BaardBouvet, line 106: The global will always support subsets, so this is irrelevant.

BaardBouvet (changes requested):

JonasAls89:
Trying to sum up on the 50 plus threads about this. Let me know if something should be restated.

patterns.rst
```diff
@@ -101,6 +101,14 @@ Optimistic locking
 ------------------
 Should be added via an external transform and then two hash values should be compared. In case of difference, discard entity.
 
+Only process the relevant entities from a global using subset
+-------------------------------------------------------------
+You should use the :ref:`subset <dataset_source>` property to pick subsets of a complete dataset, i.e: after a merge source has been used upstream. 
+
+.. note::
+
+  This pattern is sensitive to multiple versions of the entity id in one complete pipe run, i.e: if entities move between subsets, then you need to deal with these moves, as they will create old entity versions, which will be deletes.
```
grove, line 110: I don’ think we should call it move between subsets.  An entity can be part of many subsets at the same time. It is more about an entity being added to a subset and then later removed from it. Just the same was as an entity is added to dataset and then later removed from it.

patterns.rst
```diff
@@ -101,6 +101,12 @@ Optimistic locking
 ------------------
 Should be added via an external transform and then two hash values should be compared. In case of difference, discard entity.
 
+A “Best Effort” to extract entities using create and subset
+-----------------------------------------------------------
+If you use the :ref:`create <dtl-transforms>` and :ref:`subset <dataset_source>` properties to extract entities from a dataset and you do a full sync for deletion tracking, then subset in the source will still create entities that are not in the latest versions of that subset. 
+
+Therefore, you should materialise the subset in a separate dataset before “picking” your derived entities using subset to minimise the risk of creating entities that are not in the latest versions of your subset.
```
BaardBouvet, line 108: When thinking about it, why would anyone then use subset to extract entities? The pattern we should describe is "How to extract entities", which is about using create and full scan to remove orphans. We can mention that subsets can't be used for this, and the solution is to just use filtering instead.

BaardBouvet (changes requested):

patterns.rst
JonasAls89: Implemented

patterns.rst
```diff
@@ -101,6 +101,12 @@ Optimistic locking
 ------------------
 Should be added via an external transform and then two hash values should be compared. In case of difference, discard entity.
 
+How to extract entities
+-----------------------
+In general, using :ref:`subset <dataset_source>` to extract entities is straight forward. Albeit, edge cases do exist. In example, if you use the :ref:`create <dtl-transforms>` and :ref:`subset <dataset_source>` properties to extract entities from a dataset following a full scan for deletion tracking, then subset in the source will still create entities that are not in the latest versions of that subset.
```
BaardBouvet, line 106: The only way to extract entities is to use create, so this still needs a rewrite I think.

BaardBouvet (changes requested):

patterns.rst
JonasAls89: Changed it.

BaardBouvet (approved):

PR #625, opened 2022-03-21
Fix syntax example formatting

ga-hegsvold:


DTLReferenceGuide.rst
```diff
@@ -21,15 +21,17 @@ DTL consists of 'functions' that can pick and transform data and 'hops' that can
 Syntax
 ------
 
-DTL uses a JSON syntax to describe the transforms to perform. This allows us to define it directly in the pipe's configuration, which is also written in JSON. In general, DTL uses functions over keywords and as such there are just a few terms that are baked into the language.
```
BaardBouvet, line 24: Any reason to do this line break change? Makes it harder to review.

ga-hegsvold: Thought it made it easier to read when editing. I like starting sentences on a new line. Maybe just my OCD. ;-)

BaardBouvet: As a rule of thumb, never mix reformatting with real changes in the same commit. And in this case, the practice in docs is to not have sentences on a new line, so we should not do this reformatting without everyone agreeing on changing this practice.

ga-hegsvold: Reformatting reverted.
The other sections in this file have sensible line breaks.
It's just the introduction that has these wide paragraphs.

grove (approved):

PR #634, opened 2022-03-24
Update json listings with consistent indentation

ga-hegsvold:


BaardBouvet (changes requested):
I don't see how this makes the document any more consistent. Rst doesn't say how many whitespaces a code block should be indented, and the rest of the document uses a mix of 1, 2 and 3.

BaardBouvet:
Sounds good! 🙂

PR #637, opened 2022-03-25
Improved introduction

djalil-bouhachi:
Improved introduction and included something about synchronization.

intro


index-whatis.rst
```diff
@@ -1,15 +1,15 @@
...
 
-Sesam is a master data hub that simplifies the process of making
-up-to-date master data available in a data platform architecture. Sesam
-enables rapid development of solutions that can be effectively
-managed on top of a data platform. 
+Sesam is a Master Data Hub  that simplifies the process of making up-to-date master data available in a data platform architecture. 
 
-Sesam is a data platform component that collects data from source systems and connects all representations of an object across a variety of source systems in a continuously updated
-and dynamic `data model <https://en.wikipedia.org/wiki/Data_model>`_, and makes the data easily accessible inside the data platform.
+The updated data within the Sesam Data Hub is not only consumable by analytical components in the data platform, but is also synchronized back to the source systems to ensure data consistency both in analysis and systems that produce the data.
```
TarjeiSkrede, line 6: To me, just writing 'Data' instead of 'The updated data' feels better. Would probably use 'a Sesam Data Hub' instead of 'the Sesam Data Hub' too. Most important to me, this whole sentence feels like it's saying that the data will automatically be updated back to the source systems when using Sesam.

djalil-bouhachi: We want to express that synchronization of updated data across relevant systems is one of the most important benefits of using sesam. 

How about this:
The updated data within Sesam Data Hub is not only consumable by analytical components in the data platform, but **can** also be synchronized back to relevant source systems to ensure data consistency across the whole platform.

TarjeiSkrede: I understand what you want to express, the sentence just implied that it would happen automatically. The updated sentence feels better.

BaardBouvet (approved):

PR #639, opened 2022-03-25
Add skeleton for data as a product

BaardBouvet:


concepts.rst
```diff
@@ -0,0 +1,30 @@
...
+.. _whatis-global:
+
+Global
+------
+
+A global is a collection of merged data types that can be mapped to a global namespace.
+
+.. _whatis-datatype:
+
+Data type
```
djalil-bouhachi, line 13: I still think "data type" can be misleading. But I would at least consider writing it in one word "datatype".

grove (approved):

PR #662, opened 2022-04-04
Changed font to Roboto

djalil-bouhachi:
fonts


BaardBouvet:
Note to self: This is the font our new graphical profile uses (sesam.io, etc), it's not the same as the Management Studio yet.

BaardBouvet (approved):

PR #687, opened 2022-04-18
[IS-12403] Hide breadcrumbs on specific pages

ifarshgar:


BaardBouvet (dismissed):
IMHO this is a pretty bad hack and not something we should merge.

djalil-bouhachi (approved):

PR #705, opened 2022-05-02
Added docker-compose.yaml to .gitignore

kruand-sesam:
For people using docker to build the docs locally. 
Example of docker-compose file:
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ _build
 /images/Globals.pptx
 **/.DS_Store
 .vscode/settings.json
+docker-compose.yaml


BaardBouvet (changes requested):
Adding docker-compose.yaml to .gitignore doesn't make much sense and will just confuse anyone when adding that file to the repo. If you want to have local files that you don't check in, just don't add that to the Git index when you make commits.

kruand-sesam:
I don't see how it will confuse people when it is defined in .gitignore - But OK you're the boss ;)

BaardBouvet:
The convention, I think, is to add stuff to .gitignore when stuff is produced by tools and build process, typically temporary files. Can't you solve the problem here by just putting the docker compose file somewhere else? The confusion is not that it's in the .gitignore, but _why_ it's added to .gitignore. A comment in the .gitignore would make that easier.

BaardBouvet:
As an example, the first line in the .gitignore is a good example on how to not do this.

kruand-sesam:
I see your point Baard and I agree - Closing the PR :)

PR #724, opened 2022-05-16
IS-12466: Connect tutorial

ga-hegsvold:
- Fleshed out Connect tutorial
- In Collect config: changed from using "add_namespaces" to "namespaced_identifiers" to avoid potential duplication of namespaces on ``_id`` in case service.metadata already has "namespaced_indetifiers":true.

### DIFF
--- a/tutorial-getting-started-connect.rst
+++ b/tutorial-getting-started-connect.rst
@@ -1,32 +1,84 @@
-:orphan:
-
 .. _tutorial_getting_started_connect:
 
 Connect data
 ============
 
-In this tutorial we will look closer into how to...
+.. In this phase we will find sameness from our two sources and merge them into one.
+
+In this phase we will see how the same companies from different sources can be merged for a more complete perspective of the prespective companies.
+We will also get a small taste of master data management (MDM) mechanisms in Sesam.
 
 .. admonition::  Objectives:
-   
-    After you complete this tutorial you would have learned the following:
 
-    #. Objective
-    #. Objective
-    #. Objective
+    After you complete this tutorial you will have learned the following:
 
+    #. How to merge entities from different sources
+    #. How to define global properties
 
 .. admonition:: Prerequisites
 
-  Before starting on this tutorial we suggest you...
-    - Completed `Collect data tutorial <tutorial-getting-started-collect>`_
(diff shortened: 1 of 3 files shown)


ga-hegsvold:
Jonas raised a question regarding the use of `"namespaced_identifiers": false` vs `"add_namespaces": false` + `["add", "::_id", ...]` in the collect pipes to avoid adding namespaces to `_id`.
Which one is more correct?
We agreed to discuss it on the upcoming status meeting.
Some more relevant info here: https://docs.sesam.io/configuration-pipes.html?#id2

FYI @erik-leven

tutorial-getting-started-connect.rst
```diff
@@ -1,32 +1,103 @@
...
+          ]
+        ]
+      }
+    },
+    "metadata": {
+      "global": true
+    }
+  }
+
+Important to note here is that we fetch companies from *different sources* (Enhentsregisteret and HubSpot),
```
erik-leven, line 85: I think this whole section could be stripped down. The first sentence is nice to have, but I don't think we need to talk about $ids, rdf:types or coalesce. Could we maybe use a simpler language here? Something about how we create best versions of the data (without explaining how)?

ga-hegsvold: How about changing this:
"Also notice how ``$ids`` and ``rdf:type`` have data from both sources, meaning we don't lose data. We aggregate and preserve from all sources."

To:
"Also notice that data from all sources are preserved and aggregated in the result."

---

If we want to keep MDM in the intro I'm struggling with skipping to point out that ``coalesce`` is one the mechanisms we use in Sesam for MDM. The sentence can probably be refined though.

I think it's important that we describe at least a minimum of *why* we do what we do, so that it's not just an exercise in copy-pasting configs.

How about changing this:
"We use ``coalesce`` to determine the trust we have in each source, the first source being the one we trust most. This is one of the mechanisms Sesam provides for MDM."

To:
"We use ``coalesce`` to determine the source trust order and is one of the mechanisms provided by Sesam for MDM."

erik-leven (approved):

djalil-bouhachi (approved):

PR #737, opened 2022-05-20
IS-12464: Cleanup Collect tutorial

ga-hegsvold:


tutorial-getting-started-collect.rst
```diff
@@ -22,27 +22,23 @@ After having succesfully connected to these providers you will create inbound pi
     #. To set up a `test account <https://legacydocs.hubspot.com/docs/faq/how-do-i-create-a-test-account>`_
     #. To aquire an `API key <https://knowledge.hubspot.com/integrations/how-do-i-get-my-hubspot-api-key>`_
     
-|
-|
-
 Import data to HubSpot
 ^^^^^^^^^^^^^^^^^^^^^^
-In order to be able to import data to Sesam we first need to make sure that HubSpot contains data we can later import import. Therefore, the first step is to populate HubSpot with some data by following the steps below:
+In order to import data into Sesam we first need to make sure that HubSpot contains the data we want to import. Therefore, the first step is to populate HubSpot with some data by following the steps below:
```
erik-leven, line 27: contains -> contain

tutorial-getting-started-collect.rst
```diff
@@ -22,27 +22,23 @@ After having succesfully connected to these providers you will create inbound pi
...
+#. Log into HubSpot and navigate to your **Contacts** > **Companies** section
+#. Click **Import** on the right hand side of the page
+#. Click **Start an import** and select **File from computer** and click **Next**
+#. Select **Multiple files with associations** and click **Next**
+#. Select **Companies** and **Contacts** and click **Next** 
+#. Upload ``learn-hubspot-company.csv`` under **Company** and ``learn-hubspot-contacts.csv`` under **Contacts** (don't click on the "This file includes a ... column" boxes) and click **Next**
+#. Select **Company ID** as common column header and select **Company** as the object it is unique for and click **Next**
+#. Select **Don't import data in unmapped column** and click **Next**
+#. Select **Don't import data in unmapped column** again and click **Next**
+#. Agree to the terms and click **Finish import**
```
erik-leven, line 39: I prefer the old version here (Finish the import). I don't mind the bold text, but the text in HubSpot associated with "agreeing to the terms" is a bit iffy

tutorial-getting-started-collect.rst
```diff
@@ -154,18 +145,17 @@ The first inbound pipe we want to work on is the pipe that connects to our ``Hub
       "namespaced_identifiers": false
     }
 
-
-
 The Enhetsregisteret inbound pipe
 *********************************
 
-The last thing to do in this tutorial is to create the inbound pipe for Enhetsregisteret. We want to pull in the ``enhetsregisteret`` datatype from the provider. Again, follow the below steps to create your inbound pipe ``enhetsregisteret-company-collect``:
+The last thing to do in this tutorial is to create a pipe that connects to our ``enhetsregisteret`` system. We want to pull in the company data from Enhetsregisteret.
```
erik-leven, line 151: that last sentence feels a bit awkward?

tutorial-getting-started-collect.rst
ga-hegsvold: HubSpot is singular ("it"), no? So I think "contains" is correct.

tutorial-getting-started-collect.rst
ga-hegsvold: Yeah, but having to click that checkbox with "terms" threw me off so I thought it'd be good to include it there to reassure the user that it's ok to agree. If there is a consensus to keep the old version, I'm ok with that.

tutorial-getting-started-collect.rst
erik-leven: you're right, I'm dumb :D

tutorial-getting-started-collect.rst
erik-leven: Thing is, the data the users will upload is "kinda" from a third paty (us), but they are copies of the example data already in HubSpot. 

Then again, I assume "finish the import" and "Agree to the terms and click **Finish import**" kinda says the same thing in terms of liability for us.

Let's keep your version for now

tutorial-getting-started-collect.rst
ga-hegsvold: Yeah, I struggled a bit with this one. Tried to avoid using "datatype" and keep it similar to the HubSpot section above. Also thought about phrasing it in a way that would convey to the user that we want to pull company data here as well, same as from HubSpot, but fell a bit short.
Suggestions welcome!

erik-leven: What about:

The last thing to do is to pull in the company data from Enhetsregisteret by creating a pipe that connects to our ``enhetsregisteret`` system.

ga-hegsvold: I like it!

ga-hegsvold: Fixed.

erik-leven (approved):

PR #751, opened 2022-06-11
IS-12486 : Adding REST inbound pipe

JonasAls89:


learn-collect-inbound-pipes.rst
```diff
@@ -26,4 +26,27 @@ In addition, there are some best practices related to inbound pipes that we will
...
     - :doc:`What is Sesam <index-whatis>`
     - :doc:`Building Blocks in Sesam <developer-guide>`    
-    - :ref:`Inbound pipes <concepts-pipes>`
\ No newline at end of file
+    - :ref:`Inbound pipes <concepts-pipes>`
+
+.. toctree::
+   :maxdepth: 1
+
+   REST inbound pipes <tutorial-rest-inbound-pipes>
```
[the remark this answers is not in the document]
JonasAls89: I think that anchor is correct. It's the anchor at the top of tutorial-inbound-pipes-rest.rst, just above the title.

tutorial-inbound-pipes-rest.rst
```diff
@@ -4,4 +4,53 @@
...
+
+  - Created an inbound pipe connected a REST source system
+  - Looked at the dataset created by running your inbound pipe
+  - Learned about best practices associated with inbound pipes
+
+.. admonition:: Prerequisites
+
+  #. A :ref:`HubSpot REST source system <tutorial-rest-source-system>`.
+
+Sesam ensures that data residing in your external systems are retained in its raw state, or near to, in inbound pipes. As such, inbound pipes should retain data integrity. This provides a set of :ref:`advantages <collect-advantages>` in Sesam that will i.e. allow for you to repurpose a flow of data from an inbound pipe easily as data moves through Sesam towards egress. In addition, by retaining raw state of data in inbound pipes Sesam can focus on integrating data from external systems, as opposed to the external systems themselves. This is primarily because you will have replicated the shape of data in your external system.
```
erik-leven, line 21: We do not ensure this. It's best practice, but nothing in Sesam itself ensures it.

Also, I've been thinking about the language itself. Can we make it simpler? We want to make it as understandable as possible. This sounds a little to academic if you ask me.

tutorial-inbound-pipes-rest.rst
```diff
@@ -4,4 +4,53 @@
...
+
```
erik-leven, line 22: Everything below here looks a lot like the getting started tutorials. Can we add some more "meat"? 
I woul dbe nice if we could show an example of the data coming from the API and an example of the output data. That way we can also argue for why we're using "id_expression", "payload_property" and "properties". It is not very clear to me right now. Everything that's not given should be argued for and shown why we have used it.

tutorial-inbound-pipes-rest.rst
```diff
@@ -4,4 +4,53 @@
...
+	    "type": "rest",
+	    "system": "hubspot",
+	    "id_expression": "{{ id }}",
+	    "operation": "get",
+	    "payload_property": "results",
+	    "properties": {
+	      "url": "contacts?properties=city,company,email,firstname,jobtitle,lastname,state,website,twitterhandle&associations=companies"
+	    }
+	  },
+	  "namespaced_identifiers": false
```
erik-leven, line 46: Why namespaced_identifiers false?

tutorial-inbound-pipes-rest.rst
```diff
@@ -4,4 +4,53 @@
...
+	}
+
+When done you should have 2 entities in the output of your inbound pipe.
```
erik-leven, line 49: It looks like the syntethic entities GAH created is not included in that case?

tutorial-inbound-pipes-rest.rst
```diff
@@ -4,4 +4,53 @@
...
+
+.. note::
+
+  Best practice in Sesam is to always use :ref:`continuation support <continuation_support>` in your inbound pipe when the system you are working on supports it.
```
erik-leven, line 53: Can we explain this further? Why are we not using it in this pipe? Should maybe be in the introduction of the tutorial?

tutorial-inbound-pipes-rest.rst
```diff
@@ -4,4 +4,53 @@
...
+
+After having successfully created your inbound pipe, you are now ready to move onto the next tutorial in this guide, or look at the enrich guide, where you will learn to apply Sesam's :ref:`semantic enrichment <enrich>.
```
erik-leven, line 55: Talk about this during the meeting later today

tutorial-inbound-pipes-rest.rst
```diff
@@ -4,4 +4,53 @@
 REST - Inbound Pipes
 --------------------
 
-coming soon....
+In this tutorial you will add an inbound pipe connected to your REST source system: ``hubspot``.
```
erik-leven, line 7: Should we add something about the fact that this is a rest source without continuation support?
Should also reflect in the tutorial name (and .rst name?)

learn-collect-inbound-pipes.rst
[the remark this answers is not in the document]
erik-leven: Is there a reason why the document has an anchor at the top? This document can simply be called by referring to the doc name (tutorial-inbound-pipes-rest). Anchors at the top of documents aren't really needed since each document has an anchor already associated with the document itself. 

Also, it's confusing that the anchor uses a different nameing convention than the document itself

tutorial-inbound-pipes-rest.rst
JonasAls89: I'll rephrase, nicely spotted. 

To me, I like this form of conveying and writing up of things. Let's discuss this all of us?

learn-collect-inbound-pipes.rst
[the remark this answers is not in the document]
erik-leven: Yes, or atleast, Geir Ove kommented that for me a couple of months ago.

tutorial-inbound-pipes-rest.rst
JonasAls89: In accordance to what @ga-hegsvold found when writing getting started. This is what we agreed upon.

tutorial-inbound-pipes-rest.rst
erik-leven: Agreed, it's going on the agenda for today's meeting

tutorial-inbound-pipes-rest.rst
erik-leven: Sorry, I was in a hurry. Poorly written by me. The point was, should we include why we're doing that in the tutorial?

tutorial-inbound-pipes-rest.rst
JonasAls89: I can do that yes.

tutorial-inbound-pipes-rest.rst
JonasAls89: I could add that when I talk more about the other properties, as requested in an earlier comment. Sure. 👍

tutorial-inbound-pipes-rest.rst
```diff
@@ -42,13 +56,13 @@ We created a template to get you started in Sesam. Follow these steps to add and
...
+When done you should have 100 entities in the output of your inbound pipe. As you can see in the **Output** tab, these is no namespace added to any of your properties. This is due to the above property ``namespaced_identifiers: false``. This ensures that Sesam's semantic enrichment is not applied to data at this stage of your synchronization. 
 
 Extending on some of the other properties used in the above configuration the ``id_expression`` is used to tell Sesam which payload property should be used as ``_id`` for each entity produced when running the pipe whilst ``payload_property`` states which property in the HubSpot payload data should be parsed from. Finally, the ``properties`` property allows for you to state which datatype should be fetched from HubSpot in addition to properties and associations that should be retained in the payload for this specific datatype. This is done via the use of `query parameters <https://branch.io/glossary/query-parameters/>`_ in this particular use-case.
```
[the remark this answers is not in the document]
erik-leven: Also, we should start by generic explanation before this particular case. As an example: the "properties" attribute do in this case state which properties we want to import from HubSpot, but only because the API works that way. Not necessarily because that's what the "properties" attribute always does.

tutorial-source-systems-rest.rst
```diff
@@ -52,7 +50,7 @@ We created a template to get you started in Sesam. Follow these steps to add Hub
 
   Best practice in Sesam is to always use :ref:`continuation support <continuation_support>` when the system you are working on supports it.
 
-After having successfully created your REST system, you are now ready to move onto the next tutorial in this guide, or look at the :ref:`REST - Inbound Pipes <tutorial-rest-inbound-pipes>` where you will import the datatype ``contact`` from your recently created HubSpot REST system. 
+After having successfully created your REST system, you are now ready to move onto the next tutorial in this guide, or look at the :doc:`REST - Inbound Pipes <tutorial-inbound-pipes-rest>` where you will import the datatype ``contact`` from your recently created HubSpot REST system. 
 
 .. hint::
```
djalil-bouhachi, line 56: Update link to properties of REST source https://docs.sesam.io/configuration-sources-rest.html?highlight=rest#properties?

JonasAls89: There is no anchor to the properties. I think we should refrain from using link refs to our own docs? It's not that robust in my opinion. What do you think?

erik-leven: If we're missing anchors to a header in docs we can just add that header

tutorial-inbound-pipes-rest.rst
```diff
@@ -1,7 +1,71 @@
...
+The ``id_expression`` is used to tell Sesam which payload property should be used as ``_id`` for each entity produced when running the pipe whilst ``payload_property`` states which property in the HubSpot payload data should be parsed from. Finally, the ``properties`` property allows you to state which datatype should be imported from HubSpot, in this case you can see that we import the "contacts" datatype. In addition, properties, associations and limit are also specified. These are what you call `query parameters <https://branch.io/glossary/query-parameters/>`_. The use of query parameters is specific to the HubSpot API and will not necessarily look the same the next time you will import data from another REST source system. For this particular use-case, the properties query parameter specifies which attributes we want to import from the HubSpot API, whilst associations define which datatype "contacts" should be associated with and limit specifies the maximum number of entities returned when making a request.
```
erik-leven, line 68: We should start by giving a generic explanation before applying it to this particular case. As an example: the "properties" attribute do indeed in this case state which properties we want to import from HubSpot, but only because the API works that way. Not necessarily because that's what the "properties" attribute always does. For anyone who reads this and wants to use "properties" to an other API, this explanation might not make sense

tutorial-source-systems-rest.rst
JonasAls89: Added. Resolving this.

tutorial-inbound-pipes-rest.rst
JonasAls89: Nope, I have not added this as of yet. Do you reckon I should include it? Should then probably be collapsable logic in the .rst file? Have you used that before @erik-leven ?

tutorial-inbound-pipes-rest.rst
JonasAls89: I think I've done that? I feel like there is a use in keeping it generic whilst also relevant, which is what I try. I reformulated and restructured after you pointed this out the first time.

tutorial-inbound-pipes-rest.rst
erik-leven: Yeah, I think we discussed this in a meeting some weeks ago. If we show the API payload it's easier to showcase the different functionalities in the REST source.

Hmm, not sure about collapsing. I guess it depends on the size of the payload. But yeah I've used it. If you look at the source code for configuration-sources.rst I've used collapsible code there under the "Continuation support for Microservices" part

erik-leven: Since this is supposed to be a practical documentation, I think it's important that we show as much as possible, specially when we're using examples. Otherwise it may be hard to translate the examples to other situations

tutorial-inbound-pipes-rest.rst
JonasAls89: I see. I think I cover that aspect here: "... The use of query parameters is specific to the HubSpot API and will not necessarily look the same the next time you will import data from another REST source system. For this particular use-case, the properties query parameter specifies which attributes we want to import" Ok?

PR #755, opened 2022-06-20
Is 12485 alt

erik-leven:
Skjermdump fra 2022-06-20 09-35-43
Skjermdump fra 2022-06-20 09-35-33


tutorial-source-system-rest.rst
```diff
@@ -0,0 +1,75 @@
...
+  - Gotten aquainted with the different properties available for the REST source system
+
+.. admonition:: Prerequisites
+  
+  Before starting on this tutorial we suggest you read the documentation section regarding :doc:`REST system <configuration-systems-rest>`.
+
+  You will also need to create a HubSpot account and aquire an API key. You can see a step-by-step guide in the prerequisites of the *Getting started* guide :doc:`Create your first synchronization <guide-create-first-synchronization>`
+
+
+Creating the full url
```
ga-hegsvold, line 21: I recommend we write "url" as "URL". Reading "urls" feels weird.

Should we also say something about what is needed for a REST system to work before we dive into how to creating the URL? Just a very short summary/one-liner?
Something like: For REST systems to work we need to define a request URL and which HTTP operations to support (GET, PUT, etc).

tutorial-source-system-rest.rst
```diff
@@ -0,0 +1,75 @@
...
+---------------------
+
+When creating a REST system in Sesam there are several ways one can define the url to which we wish to send a request. In general you want to system to contain a base url to which you can connect endpoint specific urls depending on which API a specific pipe wishes to send a request to. In Sesam we can do this by establishing a ``url_pattern`` and let each pipe connect endpoint specific urls to this pattern.
```
ga-hegsvold, line 24: Can we simplify this a a bit?

"When creating REST systems in Sesam there are several ways to define the request URL.
In general we recommend putting the base URL in the system config and the API endpoint specific path in pipe configs.
This makes it easier to reuse systems when adding new pipes.
In Sesam we can do this by putting the base URL in ``url_pattern`` in the system config and let each pipe contain endpoint specific paths."

tutorial-source-system-rest.rst
```diff
@@ -0,0 +1,75 @@
...
+
+Creating the full url
+---------------------
+
+When creating a REST system in Sesam there are several ways one can define the url to which we wish to send a request. In general you want to system to contain a base url to which you can connect endpoint specific urls depending on which API a specific pipe wishes to send a request to. In Sesam we can do this by establishing a ``url_pattern`` and let each pipe connect endpoint specific urls to this pattern. 
+
+As you can see in the tutorial :doc:`Inbound Pipe - REST <tutorial-inbound-pipe-rest>` we can define a ``properties`` attribute in our pipe source configuration that may pass endpoint specific information to an operation in a REST system. We can connect this to the ``url_pattern`` by adding the pipe ``properties`` information to the operation url and connecting it to the ``url_pattern`` through the ``%s`` placeholder. 
+
+
+Creating the system
```
ga-hegsvold, line 29: I think it's helpful to the reader if we include the system type in the heading: "Creating the REST system"
Easier to catch when scanning through the text.

ga-hegsvold (changes requested):

tutorial-source-system-rest.rst
```diff
@@ -0,0 +1,75 @@
...
+Source System - REST
+--------------------
+
+In this tutorial we will add a REST source system to your Sesam subscription. The external system we will be working with is `HubSpot <https://www.hubspot.com/>`_ and we will connect a REST system to their APIs.
+
+.. admonition:: Objectives
+
+  After you complete this tutorial you will have:
+
+  - Successfully create a REST source system in your Sesam subscription
```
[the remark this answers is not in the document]
erik-leven: fixed

tutorial-source-system-rest.rst
erik-leven: Yeah, I thought about that. I guess the balance here is to know what we should explain here and what should be explained in docs. We can have a talk about this later at the meeting

tutorial-source-system-rest.rst
erik-leven: I like. Fixed

tutorial-source-system-rest.rst
erik-leven: agreed. Fixed

tutorial-source-system-rest.rst
```diff
@@ -0,0 +1,77 @@
...
+
+  You will also need to create a HubSpot account and aquire an API key. You can see a step-by-step guide in the prerequisites of the *Getting started* guide :doc:`Create your first synchronization <guide-create-first-synchronization>`
+
+
+Creating the full url
+---------------------
+
+When creating REST systems in Sesam there are several ways to define the request URL. In general we recommend putting the base URL in the system config, and the API endpoint specific path in pipe configs. This makes it easier to reuse systems when adding new pipes.
+
+In Sesam we can do this by putting the base URL in the ``url_pattern`` attribute in the system config and let each pipe contain endpoint specific paths in the.
```
ga-hegsvold, line 26: This sentence ends a bit abrupt ;-) "in the" can be removed.

Btw, are we switching to using the term "attribute" instead of "property"?

ga-hegsvold (changes requested):

tutorial-source-system-rest.rst
erik-leven: The attribute/property is just me switching things up. But let's add it to the meeting later and decide how strict we're gonna be

erik-leven: first comment fixed

ga-hegsvold: My opinion is we should use "property" consistent when writing about properties in Sesam.
XML uses "attribute" to describe the same, so we can use "attribute" when writing about the XML equivalents to Sesam's properties.

PR #810, opened 2022-09-07
Update global-datasets.rst

JonasAls89:
Refining this feature.

Obs. The last picture depicting Sesam with global-datasets is not ideal. It conveys the point the writer wanted to convey, but it's highly biased since the number of inbound pipes just have decreased a lot to show the point. There should be no difference in number of inbound pipes in either of the two pictures.

### DIFF
--- a/features/global-datasets.rst
+++ b/features/global-datasets.rst
@@ -3,19 +3,20 @@
 Global datasets
 ===============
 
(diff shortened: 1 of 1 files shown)


djalil-bouhachi:
@JonasAls89 can we use examples from the baseline (HubSpot)?

JonasAls89:
> @JonasAls89 can we use examples from the baseline (HubSpot)?

I think it shows the intend of the feature fairly well. I don't think we necessarily need to use HubSpot everywhere. If you feel strongly about it, I can change one of datasets used though. 👍

djalil-bouhachi:
@JonasAls89 I think we should reference the baseline as long as it is possible.

djalil-bouhachi (approved):

PR #821, opened 2022-09-14
making default optional explicit

JonasAls89:


BaardBouvet (changes requested):
I would drop the optional word from the first sentence as well.

grove (approved):

erik-leven (approved):

BaardBouvet (approved):

PR #832, opened 2022-09-23
added id text

erik-leven:
Skjermbilde 2022-09-23 120227

### DIFF
--- a/documentation/data-synchronization/insert-pattern.rst
+++ b/documentation/data-synchronization/insert-pattern.rst
@@ -39,7 +39,17 @@ Per default Sesam will pass entities with ``"_deleted": true`` through all trans
 Connect mapping data
 ^^^^^^^^^^^^^^^^^^^^
 
-Once an insert if performed, we need to store both the insert ```_id`` and the original source entity's ``_id`` to the sink dataset. This, together with the :ref:`capture response with transform pattern<capture_response_with_transform>`, will allow us to connect these two entities in the corresponding global pipe, which ensures a fully connected data flow. In the case where no other metadata can act as merge critera, this mapping is the only way to connect inserted entites with other corresponding entrys from other source systems.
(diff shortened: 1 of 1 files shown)


documentation/data-synchronization/insert-pattern.rst
```diff
@@ -39,7 +39,17 @@ Per default Sesam will pass entities with ``"_deleted": true`` through all trans
 Connect mapping data
 ^^^^^^^^^^^^^^^^^^^^
 
-Once an insert if performed, we need to store both the insert ```_id`` and the original source entity's ``_id`` to the sink dataset. This, together with the :ref:`capture response with transform pattern<capture_response_with_transform>`, will allow us to connect these two entities in the corresponding global pipe, which ensures a fully connected data flow. In the case where no other metadata can act as merge critera, this mapping is the only way to connect inserted entites with other corresponding entrys from other source systems.
+Once an entity is inserted into the external system, we generally want to collect this entity back to Sesam in order to :ref: `merge <merging-feature>` it with it's cooresponding counterparts in a global dataset. In many cases the mapping between object within the same global flow is straight foreward: customer entities from different systems might share the same organization number, or person entities might share the same SSN. However, in some cases this mapping is not available.
```
grove, line 42: Typo: cooresponding
Typo: straight foreward (should have been "straightforward")

documentation/data-synchronization/insert-pattern.rst
```diff
@@ -39,7 +39,17 @@ Per default Sesam will pass entities with ``"_deleted": true`` through all trans
...
+
+In the case where the external system returns a unique identifyer for the inserted entity, such as an ``id``, we can use this to ensure mapping between the inserted entity ans the original source entity. In those cases, once an insert if performed, we need to store both the insert ```id`` and the original source entity's ``_id`` to the sink dataset. This, together with the :ref:`capture response with transform pattern <capture_response_with_transform>`, will allow us to connect these two entities in the corresponding global pipe, which ensures a fully connected data flow. In the case where no other metadata can act as merge critera, this mapping is the only way to connect inserted entites with other corresponding entrys from other source systems.
```
grove, line 44: Typo: identifyer
Typo: ans

grove (changes requested):
There are a few typos.

PR #855, opened 2022-10-20
Update configuration-sources-sql.rst

JonasAls89:
Correcting so that query is not a required property and elaborating on what SQL query is run when only using "updated_column" to enable since support.


Ole-Morten (approved):
LGTM

PR #914, opened 2022-12-20
Added API GET parameter 'compressed' to '/datasets/{dataset_id}/entities'

kruand-sesam:
For some reason this paramter is not documented.
It is very helpful for downloading a pipe dataset to be used for tools like JQ.

The file size of the downloaded dataset will be drastically smaller with the compressed=true parameter.


kruand-sesam:
Closing the PR as the paramter seems to be specific to cURL and seemingly has nothing to do with the API endpoints.

PR #924, opened 2022-12-28
IS-13673: documented the previous_body and previous_headers jinja var…

tombech:
…iables


documentation/service-configuration/pipes/configuration-sources-rest.rst
```diff
@@ -88,7 +88,10 @@ Properties
        ``payload`` and operation ``payload`` differ, then the source payload will take precedence and the
        operations payload will be ignored. This property supports the
        ``Jinja`` template (https://palletsprojects.com/p/jinja/) syntax with the named parameters
-       ``properties``, ``url``, ``request_params``, ``since`` and ``headers`` available to the template.
+       ``properties``, ``url``, ``request_params``, ``since`` and ``headers`` available to the template.  If the
+       operation supports paging then ``previous_body`` and ``previous_headers`` is available for all page requests
+       except the first. Tip: use the Jinja "is defined" syntax for these variables to set default values for the
```
branislavjenco, line 93: Could we linked to the jinja docs for "is defined" syntax here? (and in the subsequent mentions of it)

changelog.rst
```diff
@@ -10,6 +10,7 @@ Changelog
...
 * The ``headers`` and ``params`` properties of the ``operations`` section of the :ref:`REST system <rest_system>` can now be templated using Jinja expressions.
 * The ``payload`` property of the ``operations`` section of the :ref:`REST system <rest_system>` and in the :ref:`REST source <rest_source>` , :ref:`REST transform <rest_transform>` and :ref:`REST sink <rest_sink>` configurations can now be templated using Jinja expressions.
+* Added ``previous_body`` and ``previous_headers`` named parameters to relevant "templatable" properties of the :ref:`REST system <rest_system>` and in the :ref:`REST source <rest_source>` and :ref:`REST transform <rest_transform>`. Note that these are only set for systems that supports paging, and not for the first page - use Jinja "is defined" tests in templates that use these to set default values for the first page.
```
branislavjenco, line 13: I'd say "templateable" instead of "templatable".

I'd slightly rewrite this sentence: 
"Note that these are only set for systems that supports paging, and not for the first page - use Jinja "is defined" tests in templates that use these to set default values for the first page."
to
"Note that these are only set for systems that support~s~ paging, for all pages except the first one. Use Jinja's "is defined" tests in templates that use these to set default values for the first page."

branislavjenco (approved):

PR #989, opened 2023-02-07
Add iso

djalil-bouhachi:
image

### DIFF
--- a/_templates/iso.html
+++ b/_templates/iso.html
@@ -1,0 +1,5 @@
+<div class="left-navbar-end-item-parent">
+    <ul>
+    <li.end-item><a href="#">Certified ISO 27001</a></li>
+    </ul>
+</div>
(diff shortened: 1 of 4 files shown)


_templates/iso.html
```diff
@@ -0,0 +1,5 @@
+<div class="left-navbar-end-item-parent">
+    <ul>
+    <li.end-item><a href="#">Certified ISO 27001</a></li>
```
BaardBouvet, line 3: What does this link to?

djalil-bouhachi: No links yet until we agree on it. Ref. message last week on Slack.

BaardBouvet: Then we should not mark it as a link (that doesn't work)..

BaardBouvet (approved):

PR #993, opened 2023-02-14
Added  $origin property to connect input contract

grove:


hub/documentation/data-synchronization/connectors/contract-connectors.rst
```diff
@@ -95,6 +95,10 @@ Properties
      - Boolean
      - If set to true then it is not a real delete, but rather the entity got a new `_id`. This property is produced by the merge source.
      - Optional for deletes. Disallowed for update and insert.
+   * - ``$origin``
+     - NI
```
BaardBouvet, line 99: The connector should not care about the datatype, it should be treated as an opaque value.

grove: Fixed.

BaardBouvet (approved):

PR #1130, opened 2023-11-30
Update tripletex.rst

djalil-bouhachi:
Typos


talk/systems/tripletex.rst
```diff
@@ -51,14 +51,14 @@ Customers
 For customers the following merge criteria is used:
 
 * email  
-* organizationnumner
+* organizationnumber
```
erik-leven, line 54: TT spell it with a z, so I think this has to be with a z too

erik-leven (approved):

PR #1232, opened 2024-10-02
IS-17857: Added new transform patterns section and the namespace split pattern

erik-leven:
image
image
image


parhambarazesh (approved):
lgtm (add link label later)

PR #1292, opened 2025-04-02
Ole morten patch 2

Ole-Morten:


Ole-Morten:
Updated to include the dns admin email in env variable instead of the only variable to be changed in the compose file.

heneri (approved):

PR #1298, opened 2025-04-22
Update pricing.rst

Ole-Morten:
Removed georedundant cost part because its free and added part about exceeding 1tb for multi compute


Ole-Morten:
Redundant, closing

